PR #4, opened 2019-04-09
Build sycl-xocc for check-all and move to sycl/tools

agozillon:
The problem with sycl-xocc not building directly to the bin wasn't 
related to needing to build twice.

It was related to using make check-all, I assumed that make check-all 
would build all the deployment related components of SYCL but it only 
builds a sub-section of them related to the testing infrastructure.

Since a long term goal is to have our own tests integrated with the 
existing sycl test suite, I don't see any harm in including sycl-xocc as 
a dependency for the test-suite for the time being.

This means make check-all and regular make should compile the sycl-xocc 
script to the bin appropriately.

I also moved the sycl-xocc script and directory from clang/tools to 
sycl/tools. This way it should be easier to find for people wondering 
where the script is.


keryell:
Thanks.

PR #5, opened 2019-04-12
Compiler defines, test cleanup and C++2a compilation possibly fixed

agozillon:
So for the most part this is refactoring tests, the compiler invocation the user needs to specify to compile for Xilinx FPGAs and allowing xilinx extensions to be compiled for non-xilinx devices (should just ignore them). Mostly by making some compiler generated defines based on our fsycl-xocc-device flag.

I also ran into an exception related ICE that I posted on the Intel/sycl compiler and have added my small proposed tweak to fix it here (xilinx::pipeline would cause ICE's when compiling for SPIR-V devices otherwise, plus xocc can get angsty around exceptions codes in the IR its fed).

I made the device compilation C++ standard equal the host standard that's passed in on the command line. This appears to have had the rather nice side affect of allowing C++2a to function on all the tests without any compiler errors (initial cursory diagnosis about the stubs may have been incorrect it seems!), whether this will hold up for everything now I'm unsure, but I'd advise using C++2a over C++17 now in any case as that's the end goal. Forward to the future! Or something like that...

I updated documentation based on these changes. 

I broke the changes into relevant commits that have more comprehensive comments, but these are the highlights.

Edit: Reverted the exception related change for now, more thought needs to be put into it as it was making the device break on host side exception related code (try/catch). Life appears not to be as easy as a one line fix sadly!


clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -3550,11 +3552,20 @@ void Clang::ConstructJob(Compilation &C, const JobAction &JA,
...
+
+    // Push on and enable for device side compilation
     if (Args.hasArg(options::OPT_fsycl_xocc_device)) {
       CmdArgs.push_back("-fsycl-xocc-device");
+      CmdArgs.push_back("-D__SYCL_SPIR_DEVICE__");
     }
   }
 
+  // These are defined for both the host and device compilation phases when it's
+  // a xilinx sycl device.
```
keryell: If you work at Xilinx on SYCL I expect a correct spelling of both of these words ;-)

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -3538,7 +3538,9 @@ void Clang::ConstructJob(Compilation &C, const JobAction &JA,
   if (IsSYCLDevice) {
     // We want to compile sycl kernels.
     if (types::isCXX(Input.getType()))
-      CmdArgs.push_back("-std=c++11");
+      if (Arg *A = Args.getLastArg(options::OPT_std_EQ))
```
keryell, line 3543: Oh thank you thank you! Give me the bright future! :-)

clang/lib/Frontend/CompilerInvocation.cpp
```diff
@@ -2834,9 +2834,15 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
     }
   }
 
+  Opts.SYCLIsDevice   = Args.hasArg(options::OPT_fsycl_is_device);
+  Opts.SYCLXOCCDevice = Args.hasArg(options::OPT_fsycl_xocc_device);
+  Opts.SYCLAllowFuncPtr = Args.hasFlag(options::OPT_fsycl_allow_func_ptr,
+                                  options::OPT_fno_sycl_allow_func_ptr, false);
```
keryell, line 2840: Weird indent. Perhaps
```
Opts.SYCLAllowFuncPtr =
  Args.hasFlag(options::OPT_fsycl_allow_func_ptr, options::OPT_fno_sycl_allow_func_ptr, false);
```
would fit in 80 characters?

clang/lib/Frontend/CompilerInvocation.cpp
```diff
@@ -2834,9 +2834,15 @@ static void ParseLangArgs(LangOptions &Opts, ArgList &Args, InputKind IK,
...
+
   // Set the flag to prevent the implementation from emitting device exception
   // handling code for those requiring so.
-  if ((Opts.OpenMPIsDevice && T.isNVPTX()) || Opts.OpenCLCPlusPlus) {
+  if ((Opts.OpenMPIsDevice && T.isNVPTX()) || Opts.OpenCLCPlusPlus
+    || Opts.SYCLIsDevice) {
```
keryell: Weird indent of `||`

sycl/doc/GetStartedWithSYCLCompiler.md
```diff
@@ -122,7 +122,7 @@ int main() {
 To build simple-sycl-app run following command:
 
    ```console
-   clang++ -std=c++11 -fsycl simple-sycl-app.cpp -o simple-sycl-app -lOpenCL
+   clang++ -std=c++2a -fsycl simple-sycl-app.cpp -o simple-sycl-app -lOpenCL
```
keryell, line 125: YES!!!

sycl/include/CL/sycl.hpp
```diff
@@ -38,8 +38,7 @@
 
 // These require > C++ 17 for now, but compiling all the tests with C++17 breaks
 // some Clang Unit tests involving SYCL making it difficult to test for breakage.
-#if (defined(__cplusplus) && (__cplusplus == 201703L))
-#define __SYCL_VENDOR_XILINX_EXTENSIONS__
+#if (defined(__cplusplus) && (__cplusplus >= 201703L))
 #include <CL/sycl/xilinx/fpga.hpp>
```
keryell, line 42: I am a little bit worried by the end-user interface.
For example if you have a program using a Xilinx FPGA extension but an old C++ version, it will fail because the extension does not exist, instead of the compilation to fail because of the Xilinx extension is asking for some future stuff.

sycl/include/CL/sycl/detail/scheduler/commands.cpp
```diff
@@ -14,7 +14,7 @@
 #include <CL/sycl/detail/scheduler/requirements.h>
 #include <CL/sycl/exception.hpp>
 
-#if (defined(__SYCL_VENDOR_XILINX_EXTENSIONS__))
+#if (defined(__SYCL_XILINX_ONLY__) && defined(__cplusplus) && (__cplusplus >= 201703L))
 #include <CL/sycl/xilinx/fpga/kernel_properties.hpp>
```
keryell, line 18: Probably similar comment.
Probably OK for this PR, but we need to think about a future solution at some point.

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
```diff
@@ -53,13 +53,6 @@ void pipeline(T functor) noexcept {
   functor();
 }
 
-// auto pipeline = [] (auto functor) noexcept {
```
keryell, line 56: Is there an issue to put back the modern syntax used by triSYCL instead, now you have increased the C++ version?

sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
```diff
@@ -116,18 +92,13 @@ int main(int argc, char* argv[]) {
 
     cgh.single_task<xilinx::reqd_work_group_size<1, 1, 1, krnl_sobel> >(
```
keryell: Old `> >`

sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
```diff
@@ -72,11 +52,7 @@ int main(int argc, char* argv[]) {
   constexpr auto width = 1024; // input.cols;
   constexpr auto area = height * width;
 
-#ifdef XILINX
-  selector_defines::XOCLDeviceSelector selector;
-#else
-  selector_defines::IntelDeviceSelector selector;
-#endif
+  selector_defines::CompiledForDeviceSelector selector;
   queue q { selector , property::queue::enable_profiling() };
```
keryell: ` ,` → `,` (not clear on GitHub preview... :-( )

sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
```diff
@@ -145,27 +116,23 @@ int main(int argc, char* argv[]) {
...
+          });
 
-            // capping at 0xFF means no blurring of edges when it gets
-            // converted back to a char from an int
-            sum = std::abs(magX) + std::abs(magY);
-            pixel_wb[x + y * width] = (sum > 0xFF) ? 0xFF : (char)sum;
+          // capping at 0xFF means no blurring of edges when it gets
+          // converted back to a char from an int
+          sum = std::abs(magX) + std::abs(magY);
+          pixel_wb[x + y * width] = (sum > 0xFF) ? 0xFF : (char)sum;
```
keryell: I am not sure this code is correct C++ in general, because  `char` might be signed and the you will have UB on the conversion.

After looking at the SYCL intrinsics, I prefer
```
pixel_wb[x + y * width] = min(sum, 0xFF);
```

sycl/test/xocc_tests/issue_related/integration_header_check.cpp
```diff
@@ -78,12 +77,9 @@ template <typename... Ts> class sixth_kernel;
...
 int main() {
-#ifdef XILINX
- selector_defines::XOCLDeviceSelector selector;
-#else
- selector_defines::IntelDeviceSelector selector;
-#endif
-  queue q {selector};
+  selector_defines::CompiledForDeviceSelector selector;
+  queue q { selector };
+
```
keryell: YUCK!

clang/lib/Frontend/CompilerInvocation.cpp
agozillon: Not even close unfortunately, plus that's the original indentation from the Intel implementation in this case, so I think it should be fine to leave it as is for now. Makes it more difficult to compare for differences when merging otherwise I think

sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
agozillon: good idea.

sycl/include/CL/sycl.hpp
agozillon: I think in reality the check can be removed and we just state that we only support compilation for C++17/C++2a, that will be the inevitable end result anyway as we further modify the run-time and the changes leak outside of the Xilinx extensions.

So just be explicit about not supporting earlier C++ versions. 

The alternative I can see would be to create multiple C++ interfaces that are optionally included based on the C++ implementation I suppose. Which would have to be done for anything we extend to support newer features, but that will start to look like what's already done here and doesn't seem like the ideal solution after some initial tinkering.

sycl/include/CL/sycl/detail/scheduler/commands.cpp
agozillon: The C++ restriction component can be removed I think, but the __SYCL_XILINX_ONLY__ block should probably remain as its implementation details in this case rather than some user facing piece of code. At least at the moment anyway.

sycl/include/CL/sycl.hpp
agozillon: The only real issue in doing this, is that it's going to get hard to check what tests are failing unless I go through all the SYCL tests and up their C++ std (as the Xilinx components will fail). Which I suppose will be inevitable, but cause some merge conflicts at some point I imagine.

agozillon: So I agree, I think this will/should be another PR

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
agozillon: This has nothing to do with the C++ version unfortunately, this is a limitation enforced by the SYCL specification I believe.

Pipeline isn't just a "function" in this case, it's a global variable, which the SYCL specification restricts and is currently enforced by the SYCL compiler with an error!

keryell: Interesting.
I think you found a bug in the SYCL specification.
A global `const` variable today is accepted for POD type. I think it should be extended for lambda.
The difference I think is that the lambda can be a local function, and so a local variable. I have the feeling we should allow it for capture-less lambda. Or even for POD-capturing lambda?

keryell: I created the SYCL specification issue https://gitlab.khronos.org/sycl/Specification/issues/221 to discuss this.

keryell:
Great contribution!

PR #19, opened 2019-04-17
Prefix Kernel Names with "x"

agozillon:
Seems a little arbitrary, but functions generally should start with a 
character. This was causing some problems with HW synthesis, this quick 
fix works around that for now, until we decide on a better way to name 
kernels.


keryell:
While rethinking at it, could you factorize out the prefix into an explicit `constexpr` string so that we can easily change it later without forgetting half of the work?

keryell:
After discussion, it would require some common header between SYCL Clang/LLVM and SYCL runtime that we do not have yet. So keep it like that for now...

PR #23, opened 2019-04-27
[SYCL] Removal of an old issue related test and addition of two new tests

agozillon:
Removed: issue_related/vec_math_host_conflict.cpp as it's now fixed from what I can tell so far

Added: simple_tests/vector_math.cpp to test some of the new vector stuff out
Added: issue_related/global_vs_local_lambda.cpp to demonstrate/remind of the current non-usability of global lambdas


keryell:
Thanks.

PR #24, opened 2019-04-27
[SYCL][Driver][sycl-xocc] Experimental SYCL Driver Patch, More Aligned With OpenMP and xocc script/ToolChain refactoring

agozillon:
This patch is related to the issue/discussion in Intel/SYCL: https://github.com/intel/llvm/issues/53 but also contains a significant amount of refactoring of the xocc driver implementation. 

For example, it also includes significant refactoring of the xocc ToolChain and sycl-xocc script. It largely squashes the linker and assembler stages in the script and ToolChain into one linker phase and moves ConstructJob calls from the ToolChain into the script making the script the main way to alter what occurs when building, the Toolchain's only real job is to find Xilinx tools and create a call to the script for now. 

The patch also adds an FPGA32/64 arch triple component and Xilinx vendor triple component so that we can force the compiler to use our ToolChain using fsycl-target rather than using -fsycl-xocc-device
 
The changes relating to issue 53 on the Intel SYCL repository are mainly the addition of a getOffloadingDeviceToolChain function inside the Clang Driver, which replaces hardcoded ToolChain calls/creation for CUDA/SYCL/OpenMP+NVPTX/HIP in favor of something similar to the getToolChain call that OpenMP (without NVPTX) uses. This way future users can have a separate ToolChain from the SYCL ToolChain and still have a way to redirect to it if they wish. This moves the requirement to have all device Tools redirected to and contained inside the SYCL ToolChain, making it the defacto master chain which could get a little crowded overtime. They can also have there own overload rules without having to alter the existing SYCL ToolChain overloads e.g. having to optionally specify an assembler stage based on target. However, overloading the SelectTool method inside of the SYCL ToolChain is still an option if an implementer really wants to just add new optional Tools to the existing SYCL ToolChain like we did prior to this patch.

So, in essence it should give the implementer the freedom to choose if they wish to add a completely new Offload Device ToolChain via getOffloadingDeviceToolChain (OpenMP style) or overload the SelectTool call and add additional Tools to the existing default SYCL ToolChain (MyriadToolChain style).

### DIFF
--- a/clang/lib/Driver/Driver.cpp
+++ b/clang/lib/Driver/Driver.cpp
@@ -638,7 +638,7 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
     // getOffloadingDeviceToolChain, because the device toolchain we create
     // depends on both.
     auto HIPTC = &getOffloadingDeviceToolChain(C.getInputArgs(), HIPTriple,
-                                              *HostTC, OFK);
+                                               *HostTC, OFK);
     C.addOffloadDeviceToolChain(HIPTC, OFK);
   }
 
@@ -777,7 +777,8 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
             // getOffloadingDeviceToolChain, because the device toolchain we
             // create depends on both.
             auto SYCLTC = &getOffloadingDeviceToolChain(C.getInputArgs(), TT,
-                                                     *HostTC, Action::OFK_SYCL);
+                                                        *HostTC,
+                                                        Action::OFK_SYCL);
             C.addOffloadDeviceToolChain(SYCLTC, Action::OFK_SYCL);
           }
         }
--- a/sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
+++ b/sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
@@ -10,7 +10,7 @@
   XOCC compile command:
   $ISYCL_BIN_DIR/clang++ -std=c++2a -fsycl \
     -fsycl-targets=fpga64-xilinx-unknown-sycldevice edge_detection.cpp \
-     -o edge_detection -lOpenCL `pkg-config --libs opencv`
+    -o edge_detection -lOpenCL `pkg-config --libs opencv`
 
 */
 


clang/include/clang/Driver/Driver.h
```diff
@@ -555,6 +555,22 @@ class Driver {
   const ToolChain &getToolChain(const llvm::opt::ArgList &Args,
                                 const llvm::Triple &Target) const;
 
+  /// Retrieves a ToolChain for a particular device \p Target triple
+  ///
+  /// HostTC is the host ToolChain paired with the device
```
keryell: If you go the Doxygen way, add `\param[in]`

clang/include/clang/Driver/Driver.h
```diff
@@ -555,6 +555,22 @@ class Driver {
...
+  ///
+  /// Action (e.g. OFK_Cuda/OFK_OpenMP/OFK_SYCL) is an Offloading action that is
```
keryell: Idem

clang/lib/Basic/Targets.cpp
```diff
@@ -552,6 +552,35 @@ TargetInfo *AllocateTarget(const llvm::Triple &Triple,
       return new X86_64TargetInfo(Triple, Opts);
     }
 
+
+  case llvm::Triple::fpga32: {
```
keryell, line 556: We should check with the HLS team about which triple they use.

clang/lib/Driver/Driver.cpp
```diff
@@ -619,29 +620,20 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
...
-    // Use the CUDA and host triples as the key into the ToolChains map,
-    // because the device toolchain we create depends on both.
-    auto &CudaTC = ToolChains[CudaTriple.str() + "/" + HostTriple.str()];
-    if (!CudaTC) {
-      CudaTC = llvm::make_unique<toolchains::CudaToolChain>(
-          *this, CudaTriple, *HostTC, C.getInputArgs(), OFK);
-    }
-    C.addOffloadDeviceToolChain(CudaTC.get(), OFK);
+
+    auto CudaTC = &getOffloadingDeviceToolChain(C.getInputArgs(), CudaTriple,
```
keryell, line 627: What is this `&` for?

clang/lib/Driver/Driver.cpp
```diff
@@ -619,29 +620,20 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
     DeviceTripleStr =
         HostTriple.isArch64Bit() ? "nvptx64-nvidia-cuda" : "nvptx-nvidia-cuda";
     llvm::Triple CudaTriple(DeviceTripleStr);
-    // Use the CUDA and host triples as the key into the ToolChains map,
-    // because the device toolchain we create depends on both.
```
keryell, line 623: Probably some part of the comment has to stay, the one on the CUDA triple for example.

clang/lib/Driver/Driver.cpp
```diff
@@ -619,29 +620,20 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
     DeviceTripleStr =
         HostTriple.isArch64Bit() ? "nvptx64-nvidia-cuda" : "nvptx-nvidia-cuda";
```
keryell, line 621: The CUDA width is aligned with the host one. It is not the case with SPIR above. We will have some fun in the compiler to mix for example 64-bit host and 32-bit AIE pointers for example... And we will have to do...

clang/lib/Driver/Driver.cpp
```diff
@@ -619,29 +620,20 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
...
+                                                *HostTC, OFK);
+    C.addOffloadDeviceToolChain(CudaTC, OFK);
   } else if (IsHIP) {
     const ToolChain *HostTC = C.getSingleOffloadToolChain<Action::OFK_Host>();
-    const llvm::Triple &HostTriple = HostTC->getTriple();
     StringRef DeviceTripleStr;
     auto OFK = Action::OFK_HIP;
     DeviceTripleStr = "amdgcn-amd-amdhsa";
     llvm::Triple HIPTriple(DeviceTripleStr);
-    // Use the HIP and host triples as the key into the ToolChains map,
```
keryell, line 637: Idem

clang/lib/Driver/Driver.cpp
```diff
@@ -779,15 +767,9 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
           else {
             const ToolChain *HostTC =
                 C.getSingleOffloadToolChain<Action::OFK_Host>();
-            const llvm::Triple &HostTriple = HostTC->getTriple();
-            // Use the SYCL and host triples as the key into the ToolChains map,
```
keryell, line 783: Idem.

clang/lib/Driver/Driver.cpp
```diff
@@ -5122,6 +5097,52 @@ const ToolChain &Driver::getToolChain(const ArgList &Args,
...
+          break;
+          case llvm::Triple::fpga32:
+          case llvm::Triple::fpga64:
+            TC = llvm::make_unique<toolchains::XOCCToolChain>(
+              *this, Target, HostTC, Args);
+          break;
+        }
+      break;
+      default:
+      break;
```
keryell: Curious `break`

clang/lib/Driver/Driver.cpp
```diff
@@ -5122,6 +5097,52 @@ const ToolChain &Driver::getToolChain(const ArgList &Args,
   return *TC;
 }
 
+const ToolChain &Driver::getOffloadingDeviceToolChain(const ArgList &Args,
```
keryell, line 5110: Good factorization

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -3560,21 +3560,15 @@ void Clang::ConstructJob(Compilation &C, const JobAction &JA,
...
-  if (Args.hasArg(options::OPT_fsycl_xocc_device)) {
-    CmdArgs.push_back("-fsycl-xocc-device");
-  }
+  // \todo Extend this to use getOffloadToolChains<Action::OFK_SYCL> and loop
+  // over to check for the Xilinx triple or even better make this reliant on the
+  // triple of the thing currently being compiled. To do this we would need to
+  // remove the reliance on a host side definition (__SYCL_XILINX_ONLY__)
+  // inside of InitPreprocessor.cpp
+  if (IsSYCL &&
+      C.getSingleOffloadToolChain<Action::OFK_SYCL>()
```
keryell: Move `&&` at the beginning of this line

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -13,56 +13,77 @@
...
+    LINKER_LIST="$5/$X.xo $LINKER_LIST"
 fi
+done <<< $DATA
+
+$XOCC --target $XCL_EMULATION_MODE --platform $XILINX_PLATFORM -l \
+  --save-temps -o $6 $LINKER_LIST
+
+# TODO: Cleanup temporary files we allocated in the systems tmp directory and 
+# when --save-temps is fixed for SYCL compilation optionally keep them as they
+# are quite useful when debugging
```
keryell, line 106: It is surprising this script does not grow more, if we consider all what you have removed from the C++

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -13,56 +13,77 @@
...
 
 # TODO: Pass an argument through the driver to set options like XILINX_PLATFORM, 
 # the environment variables are a quick workaround to swap boards. 
-
 usage() { echo xocc_compile: error: $2 >&2; exit $1; }
 
-[ $# -eq 0      ] && usage 1 "no xocc shell or bin"
-[ $# -eq 0      ] && usage 2 "no compilation option specified"
-[ $# -eq 0      ] && usage 3 "no data file"
+[ $# -eq 0      ] && usage 1 "no sdx bin directory containing xocc"
```
keryell: Curious repetition of `[ $# -eq 0      ]`. Use a real `if` instead?

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -13,56 +13,77 @@
...
+[ $# -eq 0      ] && usage 2 "no driver path specified"
+[ $# -eq 0      ] && usage 3 "no source file name"
+[ $# -eq 0      ] && usage 4 "no input file name"
+[ $# -eq 0      ] && usage 5 "no temporary directory specified"
+[ $# -eq 0      ] && usage 6 "no final output file specified"
+
+# Hardcoded path from the bin directory that xocc is stored in, works for now 
+# but a future TODO might be to allow this to be optionally specified as an 
+# ENV variable 
+XOCC_LIB_SPIR=$1/../lnx64/lib/libspir64-39-hls.bc
```
keryell: Put some "" everywhere, such as `"$1/../lnx64/lib/libspir64-39-hls.bc"`, just in case there are spaces in the arguments :-)

clang/lib/Basic/Targets.cpp
agozillon: I think they use fpga64-xilinx-none, at least the last time they compiled the 64 bit spir builtins

agozillon: It would be interesting to see if we can reuse any of the Target related components of HLS though

clang/include/clang/Driver/Driver.h
agozillon: this was me just copying the comment on getToolChain and modifying it a little for our use-case, probably worth keeping it similar

clang/lib/Driver/Driver.cpp
agozillon: I believe I moved it inside of the getOffloadingToolChain function which is a more general location for it (as that's where its now indexing into the map via triple key), but I can add it again/reword it a bit

clang/lib/Driver/Driver.cpp
agozillon: Yes, I don't think Intel have implemented that either, I don't think they yet distinguish between 32/64 bit SPIR based on the host either. Perhaps something to clarify with them?

At the moment we can specify 32/64 bit SPIR using the fpga64 or fpga32 bit triple and it doesn't consider the host architecture in the equation, so as long as the user is aware of the compiled for host architecture it works for the moment I think

sycl/tools/sycl-xocc/bin/sycl-xocc
agozillon: Sort of, each construct command that is a function inside of the ToolChain is essentially a single line in the script, so it condenses what is 20-30 lines of setting up arguments into a line or two.

Merging the Assembler and Linker into a single phase also simplified things quite a bit, the original ToolChain implementation was rough around the edges with a lot of places to clean and simplify things.

Overall, I quite like being able to clean it up like this, it's also a lot simpler for people to add new commands via the script at the moment, the only real reason for anyone to modify the ToolChain is to pass in/alter the arguments to the script

clang/lib/Driver/Driver.cpp
agozillon: From my understanding and I might be wrong, the getOffloadingDeviceToolChain call passes back a reference to a ptr (unique_ptr), so this is just getting a pointer to that reference so it can be passed to addOffloadDeviceToolChain. I had to think about this for a bit while making the changes..

This is following a similar syntax to getToolChain right now, in the way it's used and implemented (can see line 691 in the patch for OMP using it).

agozillon:
I marked as resolved the things I believe I fixed in the previous commit, so that it's easier to see what may need further discussion or might not be feasible in this patch

Note: Force pushed a failed rebase attempt that ruined the history and then second force push to correctly rebase on master

clang/lib/Driver/Driver.cpp
keryell: Ah yes, I got the precedence of operator wrong.
After checking https://en.cppreference.com/w/cpp/language/operator_precedence I understand.
I parsed the expression as `(&getOffloadingDeviceToolChain)()` and so I thought `getOffloadingDeviceToolChain()` because `&getOffloadingDeviceToolChain` and `getOffloadingDeviceToolChain` are the same for functions... Alzheimer...

sycl/tools/sycl-xocc/bin/sycl-xocc
keryell: Yes, it was my motivation to push for an external script. :-)

clang/lib/Driver/ToolChains/XOCC.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+    BinaryPath = xoccsAbsolutePath.str();
+
+    StringRef xoccDir = path::parent_path(xoccsAbsolutePath);
+
+    if (path::filename(xoccDir) == "bin")
+      BinPath = xoccDir;
+
+    // TODO: Check if this assumption is correct in all installations and give
+    // environment variable specifier option or an argument to the Driver
```
keryell, line 50: You could add this assumption to the installation document.

llvm/include/llvm/ADT/Triple.h
```diff
@@ -94,7 +94,9 @@ class Triple {
     wasm64,         // WebAssembly with 64-bit pointers
     renderscript32, // 32-bit RenderScript
     renderscript64, // 64-bit RenderScript
-    LastArchType = renderscript64
+    fpga32,         // 32-bit FPGA
+    fpga64,         // 64-bit FPGA
+    LastArchType = fpga64
```
keryell: I have the feeling it used to be sorted up to a certain point.
What about moving this after `bpfeb`.
It will also remove a merge conflict on `LastArchType` in the future when other arch will be added...

llvm/include/llvm/ADT/Triple.h
```diff
@@ -211,7 +214,7 @@ class Triple {
     CoreCLR,
     Simulator,  // Simulator variants of other systems, e.g., Apple's iOS
     SYCLDevice,
-    LastEnvironmentType = Simulator
+    LastEnvironmentType = SYCLDevice
```
keryell, line 217: Was it a bug or forgotten?

sycl/doc/XilinxFPGACompilation.md
```diff
@@ -148,21 +148,22 @@ The compiler invocation for the `single_task_vector_add.cpp` example inside
...
 ## Compiler invocation differences
 
-The `-fsycl-xocc-device` compiler directive is a flag we use to force certain
-things in the compiler at the moment, like picking our device ToolChain's
-Assembler and Linker over the regular SYCL ToolChain's Linker and defining some
-environment variables that are used by the runtime. It also forces the assembler
-stage of the Clang compiler to emit LLVM-IR for the moment. In the future we
-hope to remove this and have this sort of thing defined by the device target
-instead, to be more inline with the main Intel SYCL implementation.
+By setting the `fsycl-targets` to `fpga64-xilinx-unknown-sycldevice` you're
```
keryell: I think it is better to keep the `-` in the options.

sycl/doc/XilinxFPGACompilation.md
```diff
@@ -148,21 +148,22 @@ The compiler invocation for the `single_task_vector_add.cpp` example inside
...
-environment variables that are used by the runtime. It also forces the assembler
-stage of the Clang compiler to emit LLVM-IR for the moment. In the future we
-hope to remove this and have this sort of thing defined by the device target
-instead, to be more inline with the main Intel SYCL implementation.
+By setting the `fsycl-targets` to `fpga64-xilinx-unknown-sycldevice` you're 
+telling  the compiler to use our XOCC Tools and compile the device side code 
+for Xilinx FPGA. 
+
+This hasn't been tested with mutliple `fsycl-targets` yet (e.g. offloading to 
+both a Xilinx and Intel FPGA) and is unlikely to work, so I would advise
```
keryell: Not clear who is "I" for a random reader...

sycl/doc/XilinxFPGACompilation.md
```diff
@@ -148,21 +148,22 @@ The compiler invocation for the `single_task_vector_add.cpp` example inside
...
-Assembler and Linker over the regular SYCL ToolChain's Linker and defining some
-environment variables that are used by the runtime. It also forces the assembler
-stage of the Clang compiler to emit LLVM-IR for the moment. In the future we
-hope to remove this and have this sort of thing defined by the device target
-instead, to be more inline with the main Intel SYCL implementation.
+By setting the `fsycl-targets` to `fpga64-xilinx-unknown-sycldevice` you're 
+telling  the compiler to use our XOCC Tools and compile the device side code 
+for Xilinx FPGA. 
+
+This hasn't been tested with mutliple `fsycl-targets` yet (e.g. offloading to
```
keryell: Idem

sycl/test/xocc_tests/simple_tests/internal_defines.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include "../utilities/device_selectors.hpp"
+
+using namespace cl::sycl;
+
+int main() {
+  selector_defines::CompiledForDeviceSelector selector;
+  queue q {selector};
+
+#ifdef __SYCL_SPIR_DEVICE__
+        printf("__SYCL_SPIR_DEVICE__ defined on host \n");
```
keryell: Why `printf`? First this is a C++ program and even in C you should use `puts` because there is no format...

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -13,56 +13,94 @@
...
+XOCC_LIB_SPIR="$1/../lnx64/lib/libspir64-39-hls.bc"
+
+# Perhaps not so important that the linker is the one packaged with the Clang
+# build, but it's important that Opt is so that it's packaged with the correct 
+# LLVM passes. 
+OPT="$2/opt"
+LLVM_LINK="$2/llvm-link"
+XOCC="$1/xocc"
+
+$OPT -O3 -asfix -globaldce -inSPIRation -globaldce -kernelNameGen "$4" -o \
```
keryell, line 75: Should the `-O3` come from a compiler option instead?
And how to provide `-g` or whatever other option?

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -13,56 +13,94 @@
...
 # TODO: Pass an argument through the driver to set options like XILINX_PLATFORM, 
 # the environment variables are a quick workaround to swap boards. 
-
 usage() { echo xocc_compile: error: $2 >&2; exit $1; }
 
-[ $# -eq 0      ] && usage 1 "no xocc shell or bin"
-[ $# -eq 0      ] && usage 2 "no compilation option specified"
-[ $# -eq 0      ] && usage 3 "no data file"
+if [[ -z "$1" ]]; then
+  usage 1 "no sdx bin directory containing xocc"
```
keryell, line 40: SDx?
I have the feeling that `bin` is an implementation of current SDx install ad should not me mentioned here and in the following.

llvm/include/llvm/ADT/Triple.h
agozillon: Yes, at one point it seems it was alphabetically sorted, but no longer unfortunately. But I have no problems moving it into it's alphabetical position

llvm/include/llvm/ADT/Triple.h
agozillon: I think when they added the SYCLDevice triple they forgot to update the LastEnvironmentType, so it appears to be a something that was forgotten when added. Last_Type should always point to the last element. 

In either case, I haven't encountered any actual problems with the LastEnvironmentType correctly or incorrectly pointing to the final enum element yet. But worth correcting incase something does occur and confuse someone eventually.

sycl/tools/sycl-xocc/bin/sycl-xocc
agozillon: I'm not too sure I understand what you mean, what would you prefer instead?

sycl/test/xocc_tests/simple_tests/internal_defines.cpp
agozillon: Habitual usage from OpenCL kernels

sycl/tools/sycl-xocc/bin/sycl-xocc
agozillon: I think this should come in a separate patch, there are a lot of things desirable to be passed through to the script e.g. XILINX_PLATFORM and XCL_EMULATION_MODE would be better off as arguments than environment variables. However, at the bare minimum it requires parsing and then passing through more command line arguments via the XOCC ToolChain as well as some script modifications. It would also be interesting to look at some of the other possible compiler options we could filter down.

Perhaps a little outside of the scope of this patch? And can be opened up as a new issue for a future patch?

agozillon:
Updated and resolved the review points I think I have corrected, the rest are open for more discussion if required

clang/lib/Driver/Driver.cpp
```diff
@@ -619,29 +620,26 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
...
-      HIPTC = llvm::make_unique<toolchains::HIPToolChain>(
-          *this, HIPTriple, *HostTC, C.getInputArgs());
-    }
-    C.addOffloadDeviceToolChain(HIPTC.get(), OFK);
+
+    // Use the device and host triple as the key into
+    // getOffloadingDeviceToolChain, because the device toolchain we create
+    // depends on both.
+    auto HIPTC = &getOffloadingDeviceToolChain(C.getInputArgs(), HIPTriple,
+                                              *HostTC, OFK);
```
keryell: Curious alignment

clang/lib/Driver/Driver.cpp
```diff
@@ -779,15 +773,12 @@ void Driver::CreateOffloadingDeviceToolChains(Compilation &C,
...
-            if (!SYCLTC) {
-              SYCLTC = llvm::make_unique<toolchains::SYCLToolChain>(
-                  *this, TT, *HostTC, C.getInputArgs());
-            }
-            C.addOffloadDeviceToolChain(SYCLTC.get(), Action::OFK_SYCL);
+            // Use the device and host triple as the key into
+            // getOffloadingDeviceToolChain, because the device toolchain we
+            // create depends on both.
+            auto SYCLTC = &getOffloadingDeviceToolChain(C.getInputArgs(), TT,
+                                                     *HostTC, Action::OFK_SYCL);
```
keryell: Curious alignment

sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
```diff
@@ -8,8 +8,9 @@
     edge_detection -lOpenCL `pkg-config --libs opencv`
 
   XOCC compile command:
-  $ISYCL_BIN_DIR/clang++ -std=c++2a -fsycl -fsycl-xocc-device \
-    edge_detection.cpp -o edge_detection -lOpenCL `pkg-config --libs opencv`
+  $ISYCL_BIN_DIR/clang++ -std=c++2a -fsycl \
+    -fsycl-targets=fpga64-xilinx-unknown-sycldevice edge_detection.cpp \
+     -o edge_detection -lOpenCL `pkg-config --libs opencv`
```
keryell: Curious alignment

keryell:
Great!

Issue #20, opened 2019-04-18
[SYCL] Common functionality header for InSPIRation and SYCL runtime 


agozillon:
At the moment the InSPIRation LLVM pass and SYCL runtime share some common code that should be refactored into a shared header that can be reused across the LLVM pass and runtime. This will make it easier to maintain long term. 

Current Examples are:
(0) **program_manager.cpp**: `static std::string getUniqueName(const char *KernelName)`
(0) **InSPIRation.cpp**: `void setUniqueName(Function &F)`
(1) **kernel_properties.hpp**: `static std::vector<size_t> get_reqd_work_group_size(std::string mangledKernelName)`
(1) **InSPIRation.cpp**:  `SmallVector<llvm::Metadata *, 8> getReqdWorkGroupSize(const std::string& demangledName, LLVMContext &Ctx) `

Another option would be to rework these components in the future so that the LLVM Pass and runtime are less intrinsically linked. 

Related to discussion in: https://github.com/triSYCL/sycl/pull/19

PR #27, opened 2019-05-10
[SYCL][LLVM Pass][sycl-xocc] Enforce 0 as default buffer DDR bank assignment  

agozillon:
This is a short term fix for the problem with the Alveo u200 board and other boards that have a non-0 default DDR bank buffer assignment for kernel arguments. It enforces during compilation that the DDR bank a buffer is assigned to should be 0, so the u200 and other boards with different defaults will not be allowed to infer to another bank. This fixes XRTs complaints during hw_emu for the u200 in SDx 2018.3/XRT 2018.3.

This also fixes a hw_emu error the XRT runtime issues in its 2019.1 branch relating to buffers in the SYCL runtime not being assigned to DDR banks when they are created via OpenCL calls. As buffers seem to require DDR bank assignment on creation at some point between 2018.3 and 2019.1. 

This also turns off data flow strict for the linker phase via --xp param:compiler.hlsDataflowStrictMode=off, this fixes hw_emu compilation in SDx 2019.1 and has no affect in 2018.3 (its defaulted to off). This should be revisited at a later date to investigate further, when the error diagnostics from the data flow checker are a little bit more useful... 

I realized yesterday that an alternative more runtime oriented fix to this could be to retrieve the device that the buffer is to be connected to via the queue and then check which device it is and assign to a DDR bank based on that. For example, if it finds that the queue is to an Alveo u200 it could divert the buffer to DDR bank 1, rather than the usual default 0. This would require keeping up to date with all the boards defaults however, as I imagine the u200 isn't the only board that allocates to 1 as a default, which might not be ideal. It also comes at a runtime cost and would require maintenance for new boards, so I'm not a huge fan, but it is an alternative direction. 
 
Although, this patch does however put in place some of the infrastructure to compile for specific DDR banks provided the information can be passed to the Opt pass and puts in place a more simplistic method of passing information to the script via injection of environment variables through an opt pass generated bash script, no need to do complex bash manipulations over a data file, just generate the variables you need in C++! This same method should also be feasible for things like kernel SLR and CU assignments eventually.


llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <string>
+
+#include "llvm/SYCL/KernelPropGen.h"
+#include "llvm/ADT/StringRef.h"
+#include "llvm/IR/Module.h"
+#include "llvm/Pass.h"
+#include "llvm/IR/Function.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/raw_ostream.h"
```
keryell, line 26: Usually the `#include` as lexicographically sorted.

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+namespace {
+
+/// Retrieve the names for all kernels in the module and place them into a file
+struct KernelPropGen : public ModulePass {
+
+  static char ID; // Pass identification, replacement for typeid
+
+  KernelPropGen() : ModulePass(ID) {}
+
+  bool isKernel(const Function &F) {
```
keryell, line 51: Add a comment like (improve my Frenglish...)
```
/// Test if a function is actually a SPIR kernel
```
It could probably be a helper function somewhere in a common (tri)SYCL library.

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+          // This is to prevent mismatches between the SYCL runtime when 
+          // declaring OpenCL buffers and the precompiled kernel, XRT will 
+          // error out if there is a misamtch. Only OpenCL global memory is 
+          // assigned to a DDR bank, this includes constant as it's just 
+          // read-only global memory. 
+          // \todo When adding an explicit way for users to specify DDR banks 
+          // from the SYCL runtime this should be modified as well as the buffer 
+          // XRT extensions.
+          if (Arg.getType()->isPointerTy()
+           && (Arg.getType()->getPointerAddressSpace() == 1
```
keryell: Is there a  way to use some symbolic values instead of 1 or 3?
At least put a Doxygen comment at the end of each line like `//< global address space`

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -72,35 +72,46 @@ OPT="$2/opt"
 LLVM_LINK="$2/llvm-link"
 XOCC="$1/xocc"
```
keryell, line 109: I think that once you validated each `$1`, `$2`... you should use some good symbolic names and forget about these.
For example
```
driver_path=$2
```
or
```
driver_path_dir=$2
```
(if it is a directory) and so on.
Otherwise it hard to understand the following.

Also you can add test to check if the directory or file does exist...

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -72,35 +72,46 @@ OPT="$2/opt"
...
+# \todo Revisit removal of hlsDataflowStrictMode in the future to see if we
+# can compile for hw_emu with it on or if it outputs clearer diagnostics so 
+# we can track down the problem (when I tested this it seems that the call 
+# to log the error was possibly broken)
+ 
+$XOCC --xp param:compiler.hlsDataflowStrictMode=off $DDR_BANK_ARGS \
+  --target $XCL_EMULATION_MODE --platform $XILINX_PLATFORM  -l \
   --save-temps -o "$6" $LINKER_LIST
 
+# Irrelevant for compilation of binary, dumps useful debug information from
```
keryell: I do not understand the logic of the sentence. Lacking conjunctions somewhere?

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -363,10 +355,9 @@ struct InSPIRation : public ModulePass {
           kernelCallFuncSPIRify(F);
 
           counter = 0;
-          for (Function::arg_iterator I = F.arg_begin(), E = F.arg_end();
-                I != E; ++I) {
-              I->setName("arg_" + Twine{counter++});
-          }
+          for (auto &Arg : F.args())
```
keryell: Add comment about why.

keryell: Actually déjà vu. So create a `giveNameToArguments()` function or something like that?
And what if there are already names? Such as if it was created with a kernel functor constructor instead?
Should we just set the name if there is none already?

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -378,11 +369,6 @@ struct InSPIRation : public ModulePass {
 
           // Rename function name
           F.setName("sycl_func_" + Twine{funcCount++});
-
-          // Rename basic block name
-          // counter = 0;
-          // for (auto &B : F)
-          //   B.setName("label_" + Twine{counter++});
```
keryell, line 385: I do not remember the context. So it is now useless to rename blocks?
By the way it would be interesting to test with `-g` and/or `-O0` because it might create some havoc in `xocc`...

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return FileFD;
+  }
+
+  /// Visit all the functions of the module
+  bool runOnModule(Module &M) override {
+    SmallString<256> TDir;
+    llvm::sys::path::system_temp_directory(true, TDir);
+    std::string file = "KernelProperties_" + M.getSourceFileName();
+    llvm::sys::path::append(TDir, file);
+    llvm::sys::path::replace_extension(TDir, "sh",
```
keryell: If you are using some Bash extensions, clearer to use `.bash` instead?
But I do not know if you are using some `bash`-ism...

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+    llvm::raw_fd_ostream O(GetWriteStreamID(TDir.str()),
+                            true /*close in destructor*/);
+
+    if (O.has_error())
+      return false;
+  
+    // Generate a bash script to inject environment information containing 
+    // kernel properties that we need so we can compile. This script is called
+    // from sycl-xocc. #!/bin/bash
+    O << "#!/bin/bash\n";
```
keryell: Technically speaking, this line instructs Unix to execute this file with `/bin/bash` if it is an executable, i.e. `x` bit set...

But in your case you do not execute it independently, you just source it, so it is misleading.
Instead, I would suggest to transform you comment in a raw string that you output as a comment header into the file. :-)
You can probably do this for some other (missing? :-) ) comments of this program.

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+        kernelNames += (" \"" + F.getName() + "\" ").str();
+        
+        for (auto& Arg : F.args()) {
+          if (Arg.getType()->isPointerTy())
+          // if the argument is a pointer in the global (1) or constant (3) 
+          // address space it should be assigned to an explicit default DDR 
+          // Bank of 0 to prevent assignment to DDR banks that are not 0.
+          // This is to prevent mismatches between the SYCL runtime when 
+          // declaring OpenCL buffers and the precompiled kernel, XRT will 
+          // error out if there is a misamtch. Only OpenCL global memory is
```
keryell: Use a spell checker on the comments...

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    if (O.has_error())
+      return false;
+  
+    // Generate a bash script to inject environment information containing 
+    // kernel properties that we need so we can compile. This script is called
+    // from sycl-xocc. #!/bin/bash
+    O << "#!/bin/bash\n";
+   
+    std::string kernelNames = "";
```
keryell: Perhaps using `llvm::Twine` around is more in phase with LLVM coding rules.

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
keryell: Aligning `&&` with `Arg`

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -0,0 +1,147 @@
...
+          // declaring OpenCL buffers and the precompiled kernel, XRT will 
+          // error out if there is a misamtch. Only OpenCL global memory is 
+          // assigned to a DDR bank, this includes constant as it's just 
+          // read-only global memory. 
+          // \todo When adding an explicit way for users to specify DDR banks 
+          // from the SYCL runtime this should be modified as well as the buffer 
+          // XRT extensions.
+          if (Arg.getType()->isPointerTy()
+           && (Arg.getType()->getPointerAddressSpace() == 1 
+           || Arg.getType()->getPointerAddressSpace() == 3)) {
```
keryell: Alinging `||` with `Arg` of previous line.

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
keryell: You can even try to intermix Bash comments and strings concatenations...
Of course you need probably more Bash intermediate variables.
Bash metaprogramming in C++... :-)

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
agozillon: I think we could do that yes, although you have to compile Clang/LLVM and pass specific arguments to the compiler to have the IR generate names for parameters I believe. The default is to erase the names I believe.

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
agozillon: It seems xocc is fine without renamed block names at the moment, I have yet to require them at least, but I am always running -03 prior to passing it to xocc at the moment (although I believe triSYCL does so as well), so as you say it would be interesting to test that in the near future.

As for -g  it would be interesting however I'm not sure the runtime works with it yet (perhaps something to ask Intel). As I can compile an example for Intel with -g but when I run it I get a segfault. I suppose it doesn't prevent asking xocc to compile with -g though.

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
agozillon: It would be a useful addition long term although I don't think there is enough reusable functionality between all of the passes at the moment for it to be worth it, but I think as the passes grow that will change. But this might be coming back around to the idea of some kind of unified reusable library for the runtime and llvm passes i.e. https://github.com/triSYCL/sycl/issues/20

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
agozillon: This gave me a reason to read the manual section on llvm's string containers, something I had to do at some point.. I found them quite confusing when using them in the initial driver implementation as they don't quite interface with each other perfectly: http://llvm.org/docs/ProgrammersManual.html#string-like-containers 

Seems like both llvm::Twine and StringRef are not too useful in this case, llvm::Twine is for inline concatenation (and has no move or assignment operators, from what I can tell the assignment operator is deleted and the move isn't implemented and the compiler won't generate a default because of the guidelines on move operators) and StringRef is a reference to a constant string with no ownership so not really built for appending etc.

SmallString seems fine though, you are correct it does indeed function like a vector! Sadly though it doesn't interface particularly well with llvm::Twine (I wonder why though as it works well with StringRef) and the RHS of the += operators are llvm::Twine. So in the end the slightly nasty syntax of having to use .str() remains the same! It should be more efficient though and it's more consistent as it's not reliant on the host's std library.

It is fine to use std::string in LLVM though it seems, when you wish to maintain persistent heap memory, but in this case we don't really care about that.

agozillon:
Updated and marked as resolved the things I believe I managed to answer in the new commit (all prior comments I think)

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -295,19 +295,40 @@ struct InSPIRation : public ModulePass {
...
+  /// Test if a function is a SPIR kernel
   bool isKernel(const Function &F) {
     if (F.getCallingConv() == CallingConv::SPIR_KERNEL)
       return true;
     return false;
   }
 
+  /// Test if a function is a non-intrinsic SPIR function, indicating that it is
+  /// a user created function that the SYCL compiler has transitively generated
+  /// or one that comes from an existing library of SPIR functions (hls spir
```
keryell: HLS SPIR instead?

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -295,19 +295,40 @@ struct InSPIRation : public ModulePass {
...
         && !F.isIntrinsic())
       return true;
     return false;
   }
 
+  /// This function gives llvm::function arguments with no name
+  /// a default name e.g. arg_0, arg_1..
+  ///
+  /// Thus is because if your arguments have no name xocc will commit sepuku
+  /// when generating xml. Perhaps it's possible to move this to the Clang
```
keryell: XML

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -31,13 +31,23 @@ using namespace llvm;
...
 
+/// Enum for Address Space values, used in ASFixer pass and LLVM-SPIRV
+enum SPIRAddressSpace {
+  SPIRAS_Private,  // Address space: 0
+  SPIRAS_Global,   // Address space: 1
+  SPIRAS_Constant, // Address space: 2
+  SPIRAS_Local,    // Address space: 3
+  SPIRAS_Generic,  // Address space: 4
+};
+
```
keryell, line 42: This is clearer!

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -61,74 +71,78 @@ struct KernelPropGen : public ModulePass {
...
+    for (auto &F : M.functions()) {
       if (isKernel(F)) {
         kernelNames += (" \"" + F.getName() + "\" ").str();
-        
+
         for (auto& Arg : F.args()) {
           if (Arg.getType()->isPointerTy())
-          // if the argument is a pointer in the global (1) or constant (3) 
-          // address space it should be assigned to an explicit default DDR 
+          // if the argument is a pointer in the global or constant
```
keryell, line 96: Isn't it clearer? :-)

llvm/lib/SYCL/KernelPropGen/KernelPropGen.cpp
```diff
@@ -61,74 +71,78 @@ struct KernelPropGen : public ModulePass {
...
-           || Arg.getType()->getPointerAddressSpace() == 3)) {
-              // This currently forces a default assignment of DDR banks to 0 
-              // as some platforms have different Default DDR banks and buffers 
-              // default to DDR Bank 0. Perhaps it is possible to query the 
-              // specific platform and reassign the buffers to different default 
-              // DDR banks based on the platform. But this would require a 
-              // change for every new platform. In either case, this puts in 
-              // infrastructure to assign DDR banks at compile time for a CU 
+              && (Arg.getType()->getPointerAddressSpace() == SPIRAS_Global
+              || Arg.getType()->getPointerAddressSpace() == SPIRAS_Constant)) {
```
keryell, line 109: And safer. :-)

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -60,58 +61,86 @@ if [[ -z "$6" ]]; then
...
+  fi
+
+  LLVM_LINK="$DRIVER_PATH_DIR/llvm-link"
+  if [[ ! -f "$LLVM_LINK" ]]; then
+    usage 8 "no llvm-link executable found in $LLVM_LINK"
+  fi
+fi
 
 $OPT -O3 -asfix -globaldce -inSPIRation -globaldce -kernelPropGen "$4" -o \
   "$5/$3_kernels-optimized.bc"
```
keryell: I have the feeling that here and in the following we should have some symbolic names instead of `$3`, `$4` and `$5`

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -60,58 +61,86 @@ if [[ -z "$6" ]]; then
...
 
-# Irrelevant for compilation of binary, dumps useful debug information from 
-# the generated xcl binary into the temp directory. E.g. what args are bound 
-# to which banks, overall kernel signature etc. 
+# This step is unconnected to the generation of the final xcl binary, it dumps
+# useful debug information from the final generated xcl binary into the temp
+# directory. E.g. what args are bound to which banks, overall kernel signature
+# etc.
 if [ -z "$XILINX_XRT" ]; then
   $XILINX_XRT/bin/xclbinutil --info --input "$6" &> "$5/$6.xclbinutil.dump"
```
keryell: All the `$6` too...

agozillon:
Marked everything I think I fixed in the last commit as resolved.

keryell:
By reviewing https://github.com/triSYCL/sycl/pull/88 I looked at this.
@Ralender Perhaps an issue should be opened to track some stuff discussed in the description of this PR.

Issue #28, opened 2019-05-20
Compiler crash with named function object


j-stephan:
I am trying to compile the attached code (simple N-body) with the current versions of the compiler (up-to-date `sycl/unified/master`) and SDAccel (2018.3) and a self-compiled XRT (up-to-date `2018.3`), OS is Ubuntu 18.10. After starting the compilation the compiler crashed almost immediately with the following message:

```
$ clang++ -std=c++2a -fsycl -fsycl-targets=fpga64-xilinx-unknown-sycldevice nbody.cpp -o nbody.sw_emu -lOpenCL
Stack dump:
0.	Program arguments: /home/jan/software/sycl/bin/opt -O3 -asfix -globaldce -inSPIRation -globaldce -kernelNameGen /tmp/nbody-4bd090.o -o /tmp/nbody_kernels-optimized.bc 
1.	Running pass 'ASFixer' on module '/tmp/nbody-4bd090.o'.
 #0 0x0000562b936c72ca llvm::sys::PrintStackTrace(llvm::raw_ostream&) (/home/jan/software/sycl/bin/opt+0x20932ca)
 #1 0x0000562b936c5194 llvm::sys::RunSignalHandlers() (/home/jan/software/sycl/bin/opt+0x2091194)
 #2 0x0000562b936c5315 SignalHandler(int) (/home/jan/software/sycl/bin/opt+0x2091315)
 #3 0x00007f7f8a12fdd0 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x12dd0)
 #4 0x0000562b930e1fa3 llvm::Value::getContext() const (/home/jan/software/sycl/bin/opt+0x1aadfa3)
 #5 0x0000562b930e395a llvm::ValueHandleBase::AddToUseList() (/home/jan/software/sycl/bin/opt+0x1aaf95a)
 #6 0x0000562b93701113 llvm::ValueMap<llvm::Value const*, llvm::WeakTrackingVH, llvm::ValueMapConfig<llvm::Value const*, llvm::sys::SmartMutex<false> > >::operator[](llvm::Value const* const&) (.isra.418) (/home/jan/software/sycl/bin/opt+0x20cd113)
[...]

PR #29, opened 2019-05-21
[SYCL][sycl-xocc] Reversing -asfix and -O3 in device Opt pass chain

agozillon:
* Bug fix for https://github.com/triSYCL/sycl/issues/28


keryell:
Thanks for the quick fix!

PR #30, opened 2019-05-22
[SYCL] Required changes for a functional upstream merge

agozillon:
A lot of the changes are porting things from Scheduler 1 to 2, so that our alterations work on both. Important as Scheduler 2 is now the default, you can enable the old scheduler via -DSCHEDULER_10 when compiling a file.

Added a new minor pass that can be built on over time, it downgrades a new attribute: https://reviews.llvm.org/D57825#change-3IAnpkSikxwF so it doesn't kill xocc as its not ToT

Also reverted some changes I made to SemaSYCL.cpp to allow a wider range of namespaces, Intel added the functionality themselves. 

Probably the "biggest" change that will probably need some discussion is the changes to the SYCL (libSYCL.so) build system as I'm not too sure if its the best way forward but it works at the moment. There is an optional define of __SYCL_XILINX_ONLY__ (also used/defined by the compiler when compiling for xocc) when XRT is detected in a users environment during CMake configuration. It forces the runtime to compile certain code paths that use Xilinx extensions or the normal path based on the current platform, this is important for source files (like memory_manager.cpp) but irrelevant for the header files (like buffer_impl.hpp, still ideal to keep the details unified however) as source will be compiled into libSYCL.so and will be concrete and unchangeable by any define generated by the compiler when compiling for a specific target. This define is needed so that it's still possible to compile libSYCL.so without an XRT installation. There should be some further notes on this in the .cpp files.

As I couldn't quite merge my prior branch and show changes this patch is basically a diff applied to unified/next so the commits are squashed a bit better but might be a little out of order (shouldn't really matter and it's possible to see the original chain of commits on my sycl/unified/next branch right now).

I also need to test it all still functions after the application (I'll do so tomorrow morning as it'll take some time to build), but it shouldn't take long and this pull request will give you a head start with the review hopefully!

### DIFF
--- a/sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
+++ b/sycl/test/xocc_tests/sdaccel_ports/vision/edge_detection/edge_detection.cpp
@@ -93,6 +93,9 @@ int main(int argc, char* argv[]) {
     printf("pixel_rb count in submit: %zu \n", pixel_rb.get_count());
 
     cgh.single_task<xilinx::reqd_work_group_size<1, 1, 1, krnl_sobel>>(
+    // The reqd_work_group_size is already actually applied internally for single
+    // tasks but this showcases it's usage none the less, as it can be applied
+    // to parallel_fors with local sizes
      [=]() {
       auto gX = xilinx::partition_array<char, 9,
                 xilinx::partition::complete<0>>({-1, 0, 1, -2, 0, 2, -1, 0, 1});
@@ -100,7 +103,7 @@ int main(int argc, char* argv[]) {
       auto gY = xilinx::partition_array<char, 9,
                 xilinx::partition::complete<0>>({1, 2, 1, 0, 0, 0, -1, -2, -1});
 
-      int magX, magY, gI, pIndex, sum;
+      int magX, magY, gI, pIndex;
 
       // Simplified version of krnl_sobelfilter.cl that gives the same output
       // results however the krnl_sobelfilter.cl is more hardware optimized than
@@ -133,8 +136,8 @@ int main(int argc, char* argv[]) {
 
           // capping at 0xFF means no blurring of edges when it gets
           // converted back to a char from an int
-          sum = std::abs(magX) + std::abs(magY);
-          pixel_wb[x + y * width] = min(sum, 0xFF);
+          pixel_wb[x + y * width] = cl::sycl::min((int)(cl::sycl::abs(magX)
+                                                  + cl::sycl::abs(magY)), 0xFF);
         }
       }
     });


sycl/CMakeLists.txt
```diff
@@ -81,26 +81,37 @@ endif()
...
+#  libSYCL.so, this unfortunately isn't relevant when a user is compiling there
+#  own files, they'll still have to add -I/opt/xilinx/xrt/include/.
+#  It's only relevant to the clang build system and the libSYCL.so file that's
+#  compiled, the header files are recompiled on a per application basis as far as
+#  I'm aware.
+#  \todo We could enforce the download and installation of the XRT runtime
+#  similarly to what's done with the OpenCL headers at the moment that makes
+#  things a little easier. As it wouldn't matter if someone only wanted to build
+#  to run on another platform, they would always have XRT and it's extensions
+#  wouldn't break the runtime by being missing'
```
keryell: I am unsure this change to a double space is required here

sycl/CMakeLists.txt
```diff
@@ -81,26 +81,37 @@ endif()
...
-# I'm aware.
-# NOTE: This isn't really used at the moment as the extensions are only really
-# used in the buffer_impl.h file for now and those are walled behind Xilinx
-# IFDEF's for now. However, ideally we would move away from macro usage and
-# check the device context for a Xilinx device in use and optionally execute the
-# Xilinx Extensions rather than optionally compile them. This, however would
-# make XRT a required installation for compiling for any device not just Xilinx
-# devices, there is a related \todo in buffer_impl.hpp for this
+#  Try to find and include XRT OpenCL in the build components to allow access to
+#  Xilinx CL extensions in the source files when compiling the runtimes
```
keryell: What is CL?

sycl/CMakeLists.txt
```diff
@@ -81,26 +81,37 @@ endif()
...
+#  libSYCL.so, this unfortunately isn't relevant when a user is compiling there
+#  own files, they'll still have to add -I/opt/xilinx/xrt/include/.
+#  It's only relevant to the clang build system and the libSYCL.so file that's
```
keryell: Clang

sycl/include/CL/sycl/detail/buffer_impl.hpp
```diff
@@ -599,28 +599,33 @@ void buffer_impl<AllocatorT>::allocate(QueueImplPtr Queue,
...
-                      | CL_MEM_EXT_PTR_XILINX, ByteSize, &mext, &Error);
+    // \todo Alternatively the cl_mem_ext_ptr_t assignment of DDR banks seems to
+    // be legacy in 2019.1, it may be possible to assign buffers to kernel
+    // arguments in just the runtime and bypass the requirements for specifying
+    // appropriate DDR banks, if that makes things easier, this can be done via
+    // the alternate union form of cl_mem_ext_ptr_t:
+    // struct { // interpreted kernel arg assignment
+    // unsigned int argidx;  // Top 8 bits reserved for XCL_MEM_EXT flags
+    // void *host_ptr_;      // use as host_ptr
+    // cl_kernel kernel;
```
keryell: It probably worths to indent the content of the struct with 2 spaces...

sycl/source/detail/memory_manager.cpp
```diff
@@ -103,9 +103,58 @@ void *MemoryManager::allocateMemBuffer(ContextImplPtr TargetContext,
...
+#if (defined(__SYCL_XILINX_ONLY__))
+  // This currently enforces assignment of all buffers to DDR bank 0 via
+  // Xilinx OpenCL extensions, which we also enforce when compiling the
+  // kernels via xocc (0 is usually the default inferred space, but some get
+  // inferred to bank 1, notably Alveo U200 boards). XRT seems to have slowly
+  // gotten stricter with these assignments when compiling for hw_emu, so it's
+  // something we have to do to conform for the moment (but generally a good
+  // thing to conform as it means closer alignment to actual hardware
+  // compilation).
+  // \todo I believe a way to allow users to specify DDR bank assignments at a
```
keryell: The problem when we read the comment is that we have not idea about who is "I" owning this belief...

sycl/source/detail/memory_manager.cpp
```diff
@@ -103,9 +103,58 @@ void *MemoryManager::allocateMemBuffer(ContextImplPtr TargetContext,
...
+  // SYCL level should be the end goal here, assign a DDR bank to kernel/CU
+  // mapping via an accessor or buffer. The hard part of this is that the
+  // compiler will need access to this information when compiling the kernels,
+  // pushing this data through in a "C++ way" without modifying the SYCL
+  // compiler will be difficult I think (tying this information up as an
```
keryell: Who do really think here?

sycl/source/detail/memory_manager.cpp
```diff
@@ -103,9 +103,58 @@ void *MemoryManager::allocateMemBuffer(ContextImplPtr TargetContext,
...
+  // extra component of the accessor's could be the ideal route in this case).
+  // \todo Alternatively the cl_mem_ext_ptr_t assignment of DDR banks seems to
+  // be legacy in 2019.1, it may be possible to assign buffers to kernel
+  // arguments in just the runtime and bypass the requirements for specifying
+  // appropriate DDR banks, if that makes things easier, this can be done via
+  // the alternate union form of cl_mem_ext_ptr_t:
+  // struct { // interpreted kernel arg assignment
+  // unsigned int argidx;  // Top 8 bits reserved for XCL_MEM_EXT flags
+  // void *host_ptr_;      // use as host_ptr
+  // cl_kernel kernel;
```
keryell: See my previous comment about indent here.

sycl/source/detail/memory_manager.cpp
```diff
@@ -103,9 +103,58 @@ void *MemoryManager::allocateMemBuffer(ContextImplPtr TargetContext,
...
+  // This hasn't been tested but may be an alternative to assigning DDR banks
+  // per buffer. However, being able to specify DDR bank assignments to a
+  // kernel, is still very useful, but if this method works, perhaps we can
+  // detach the idea of assigning DDR banks via the buffer/accessor (which
+  // makes it difficult to pass information to the kernel) and instead have it
+  // as part of the kernel name via a kernel property similar to
+  // the way we handle reqd_work_group_size at the moment
+  cl_mem NewMem;
+  if (TargetContext->get_platform().get_info<info::platform::vendor>()
+      =="Xilinx") {
```
keryell: Space in '`== `'

sycl/source/detail/memory_manager.cpp
```diff
@@ -103,9 +103,58 @@ void *MemoryManager::allocateMemBuffer(ContextImplPtr TargetContext,
...
+    cl_mem_ext_ptr_t mext = {0};
+    mext.banks = 0 | XCL_MEM_TOPOLOGY;
+    mext.host_ptr = UserPtr;
```
keryell, line 145: We should come with a sensible C++ API to setup these memory properties at some point in the future instead of bit-banging the low-level `struct` everywhere...

sycl/source/detail/memory_manager.cpp
```diff
@@ -338,6 +387,14 @@ void *MemoryManager::map(SYCLMemObjT *SYCLMemObj, void *Mem, QueueImplPtr Queue,
       AccessRange[0], DepEvents.size(),
       DepEvents.empty() ? nullptr : &DepEvents[0], &OutEvent, &Error);
   CHECK_OCL_CODE(Error);
+
+  // XRT doesn't really  seem to play well with Dependent Events in
```
keryell: Too many spaces

sycl/source/detail/scheduler/commands2.cpp
```diff
@@ -366,11 +370,36 @@ cl_int ExecCGCommand::enqueueImp() {
...
+    // There is no SYCL specification restriction on this use of the attribute
+    // but perhaps it's not needed. Although, maybe there is use case for someone
+    // enforcing a reqd_work_group_size on the global range a kernel is invoked
+    // with. In either case it can be removed and all you would have to do for
+    // the case of single_task is check if the SingleTask variable is true and
+    // if it is specify a LocalWorkSize of 1,1,1 else specify nullptr (this
+    // meets the OpenCL restrictions on ReqdWorkGroupSize).
+    // This will also enforce LocalWorkSize on Intel devices if
+    // reqd_work_group_size is specified and the std is >= 17. Perhaps not
+    // ideal.
```
keryell: Otherwise you want to enforce the WG size just for performance reason even if there is no visible difference.
Also the OpenCL interoperability mode could have some kernel code using it...

sycl/test/xocc_tests/simple_tests/reqd_work_group_size.cpp
```diff
@@ -83,191 +83,214 @@ int main() {
...
+  {
+    /*
+    Invalid and breaks. While it is smaller than the specified global and
+    local_work_size of the nd_range it's not the same as the nd_range...
+
+    We don't want to enforce a reqd_work_group_size of 4x4x4 on something that's
+    already defined by an nd_range to be 8x8x8. It's possible to
+    enforce the reqd_work_group_size property over the nd_range similar to the
+    parallel_for where the local size isn't specified (for now, not sure if
+    this is what we really want to do) but your breaking the API as the
```
keryell: your → you're/you are?

sycl/source/detail/memory_manager.cpp
agozillon: The power of git blame! But yes, I get the point, although I don't think changing the sentence to non-ownership or group ownership really makes a significant difference to the readers understanding in most cases. Perhaps using initials at the start of the sentence instead may be more useful? e.g. 'AG: I believe a...'

sycl/source/detail/memory_manager.cpp
agozillon: I agree, this would be ideal and perhaps for some of the other useful OpenCL extensions. Although, once Intel remove the old scheduler (if they do) it'll only be in the one place, inside this file, I believe it'll be removed from buffer_impl.h at that point.

agozillon:
Marked everything I believe is resolved as resolved from the last review.

I also fixed somes tests for Intel GPU (id_mangle.cpp wasn't sane for any kind of parallel device and edge_detection was using a std library function to work around earlier math functions not fully working and std library calls don't seem to be appreciated!)

Fixed a bug I managed to add to the sycl-xocc script when I was altering it previously, harmless when only compiling 1 file, but would probably result in some pretty strange bugs if building multiple files in parallel. 

Also changed the functionality of reqd_work_group_size, it's now mostly just a compile time hint/request and we do some very minor tweaking in the runtime to make sure single_task has a local size of 1,1,1 rather than nullptr (to prevent OpenCL runtime errors as it's one of the restrictions)

keryell:
Thanks for the work!

PR #37, opened 2019-06-06
[SYCL] Upstream Merge Fixes and Copy/Fill Fix

agozillon:
This patch:
* Fixes some OpenCL SPIR builtins (converts them to correct mangling again)
* Separates them into a new folder and removes them from the handler's, they should now be properly invoked through the same call as the SPIRV builtins
* Adds a new test relating to the handlers copy method
* Fixes 1D copy for hw_emu (and fill), 2D/3D copy will need some further work as XRT complains, more information in the newly added test. This was done by adding some defines around the host parallel_for which was killing one of xocc's passes (no address space on one of the arguments as it's accessed via raw pointer rather than accessor). This will most likely be temporary as there is a #TODO relating to removing the parallel_for in lieu of a memory_manager solution.
* Taught the script to not kill xocc when there is no kernels found in a file

### DIFF
--- a/llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
+++ b/llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
@@ -81,10 +81,10 @@ static const std::regex matchSomeNaturalInteger {R"(\d+)"};
 // This is to give clarity to why we negate a value from the Z mangle component
 // rather than having a magical number, we have the size of the string we've
 // removed from the mangling.
-static const std::string SPIROCL("__spir_ocl_");
-static const std::string SPIRVOCL("__spirv_ocl_");
-static const std::string SPIRVOCLS("__spirv_ocl_s_");
-static const std::string SPIRVOCLU("__spirv_ocl_u_");
+static const std::string SPIROCL {"__spir_ocl_"};
+static const std::string SPIRVOCL{"__spirv_ocl_"};
+static const std::string SPIRVOCLS{"__spirv_ocl_s_"};
+static const std::string SPIRVOCLU{"__spirv_ocl_u_"};
 
 /// Transform the SYCL kernel functions into xocc SPIR-compatible kernels
 struct InSPIRation : public ModulePass {
@@ -101,7 +101,7 @@ struct InSPIRation : public ModulePass {
 
   /// Note: running this call on real SPIRV builtins is unlikely to yield a
   /// working SPIR builtin as they 1) May not be named the same/have a SPIR
-  /// equivelant 2) Are not neccessarily function calls, but possibly a magic
+  /// equivalent 2) Are not necessarily function calls, but possibly a magic
   /// variable like __spirv_BuiltInGlobalSize, something more complex would be
   /// required.
   void removePrefixFromMangling(Function &F, const std::regex Match,
@@ -409,7 +409,7 @@ struct InSPIRation : public ModulePass {
       // the builtin implementation stabilizes
       removePrefixFromMangling(*F, matchSPIRVOCLU, SPIRVOCLU);
       removePrefixFromMangling(*F, matchSPIRVOCLS, SPIRVOCLS);
-      removePrefixFromMangling(*F, matchSPIRVOCL,  SPIRVOCL);
+      removePrefixFromMangling(*F, matchSPIRVOCL, SPIRVOCL);
       removePrefixFromMangling(*F, matchSPIROCL, SPIROCL);
     }
 
--- a/sycl/include/CL/__spir/spir_vars.hpp
+++ b/sycl/include/CL/__spir/spir_vars.hpp
@@ -13,7 +13,7 @@
 // Although the __spirv_BuiltInSubgroupSize builtins at the end of the file
 // currently do not have a SPIR variation, so they will break at the moment,
 // they're included here so as not to break the compilation for the moment
-// (although it may be more desireable than a runtime error/symbol missing
+// (although it may be more desirable than a runtime error/symbol missing
 //  error)
 
 // These are converted to SPIR builtins by the inSPIRation pass.
@@ -24,7 +24,7 @@ size_t __spir_ocl_get_local_id(uint dimindx);
 size_t __spir_ocl_get_global_offset(uint dimindx);
 size_t __spir_ocl_get_group_id(uint dimindx);
 
-enum SYCLBuiltinTypes {
+enum class SYCLBuiltinTypes {
    SYCLBuiltinGlobalSize,
    SYCLBuiltinGlobalInvocationId,
    SYCLBuiltinWorkgroupSize,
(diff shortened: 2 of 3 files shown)


llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -80,6 +81,7 @@ static const std::regex matchSomeNaturalInteger {R"(\d+)"};
 // This is to give clarity to why we negate a value from the Z mangle component
 // rather than having a magical number, we have the size of the string we've
 // removed from the mangling.
+static const std::string SPIROCL("__spir_ocl_");
```
keryell: `{ }` on all the constructors to be modern...

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -91,14 +93,19 @@ struct InSPIRation : public ModulePass {
...
-                                  const std::string Namespace) {
+  /// This function currently works by checking for certain prefixes, and
+  /// removing them from the mangled name, this currently is used for
+  /// get_global_id etc. (as we forcefully prefix it with __spir_ocl_), and
+  /// the math builtins which are prefixed with __spirv_ocl_. An example is:
+  ///   Z24__spir_ocl_get_global_idj - > Z13get_global_idj
+
+  /// Note: running this call on real SPIRV builtins is unlikely to yield a
+  /// working SPIR builtin as they 1) May not be named the same/have a SPIR
+  /// equivelant 2) Are not neccessarily function calls, but possibly a magic
```
keryell: equivalent
necessarily
where is the spell checker gone?...

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -391,18 +399,18 @@ struct InSPIRation : public ModulePass {
...
-      // the decision on the upstream reflower is made though.
-      // It will check all of the manglings before settling on the base
-      // __spirv_ocl_ mangle
-      renameSPIRVIntrinsicToSPIR(*F, matchSPIRVOCLU, SPIRVOCLU);
-      renameSPIRVIntrinsicToSPIR(*F, matchSPIRVOCLS, SPIRVOCLS);
-      renameSPIRVIntrinsicToSPIR(*F, matchSPIRVOCL,  SPIRVOCL);
+      // the builtin implementation stabilizes
+      removePrefixFromMangling(*F, matchSPIRVOCLU, SPIRVOCLU);
+      removePrefixFromMangling(*F, matchSPIRVOCLS, SPIRVOCLS);
+      removePrefixFromMangling(*F, matchSPIRVOCL,  SPIRVOCL);
```
keryell: Probably the double space does not fit is the coding style.

sycl/include/CL/__spir/spir_vars.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+// These are converted to SPIR builtins by the inSPIRation pass.
+size_t __spir_ocl_get_global_size(uint dimindx);
+size_t __spir_ocl_get_local_size(uint dimindx);
+size_t __spir_ocl_get_global_id(uint dimindx);
+size_t __spir_ocl_get_local_id(uint dimindx);
+size_t __spir_ocl_get_global_offset(uint dimindx);
+size_t __spir_ocl_get_group_id(uint dimindx);
+
+enum SYCLBuiltinTypes {
```
keryell: Is there some value for not using some scoped enum instead of the old leaky C one?

sycl/include/CL/__spir/spir_vars.hpp
```diff
@@ -0,0 +1,90 @@
...
+};
+
+// Just infrastructure to use the same calls as the SPIRV builtins inside
+// the handler and then redirect to the relevant SPIR builtins instead.
+//
+// Note: There is an arguably more powerful but harder to maintain
+// implementation of this in a prior commit that could allow better
+// redirection/offloading to arbitrary builtins, if the need arises.
+template <int ID>
+size_t MapTo(SYCLBuiltinTypes builtin) {
```
keryell: It is not clear to me you need to have all this very dynamic.
What about having `constexpr MapTo` and then you can have `ID` also a normal parameter.
It looks like there is no namespace in this file. Is it on purpose?

sycl/include/CL/__spir/spir_vars.hpp
```diff
@@ -0,0 +1,90 @@
...
+DEFINE_SYCL_SPIR_CONVERTER(GlobalInvocationId)
+DEFINE_SYCL_SPIR_CONVERTER(WorkgroupSize)
+DEFINE_SYCL_SPIR_CONVERTER(LocalInvocationId)
+DEFINE_SYCL_SPIR_CONVERTER(WorkgroupId)
+DEFINE_SYCL_SPIR_CONVERTER(GlobalOffset)
+
+#undef DEFINE_SYCL_SPIR_CONVERTER
+
+// SPIRV intrinsics/builtins that we currently don't map for yet (unsure if
+// they have equivelant SPIR or HLS mappings).
```
keryell: equivalent

sycl/include/CL/__spir/spir_vars.hpp
```diff
@@ -0,0 +1,90 @@
...
+// TODO: Look into these to see if we have equivelants
```
keryell: equivalent

sycl/include/CL/sycl/handler2.hpp
```diff
@@ -951,6 +932,17 @@ class handler {
   copy(accessor<T_Src, Dims, AccessMode, AccessTarget, IsPlaceholder> Src,
        T_Dst *Dst) {
     if (MIsHost) {
+     // This makes sure we don't compile this on the device, it will kill one of
+     // the xocc hw_emu passes as it's passing a raw pointer to a kernel with no
+     // address space qualifier because we're not wrapping it in a buffer and
+     // using an accessor.
+     // We don't actually care about this parallel_for on device it's a pure
```
keryell: "since it's a pure" ?

sycl/include/CL/sycl/handler2.hpp
```diff
@@ -989,6 +981,17 @@ class handler {
        accessor<T_Dst, Dims, AccessMode, AccessTarget, IsPlaceholder> Dst) {
 
     if (MIsHost) {
+     // This makes sure we don't compile this on the device, it will kill one of
+     // the xocc hw_emu passes as it's passing a raw pointer to a kernel with no
+     // address space qualifier because we're not wrapping it in a buffer and
+     // using an accessor.
+     // We don't actually care about this parallel_for on device it's a pure
```
keryell: "since it's a pure" ?

sycl/include/CL/sycl/handler2.hpp
```diff
@@ -1042,6 +1045,17 @@ class handler {
            Dst) {
 
     if (MIsHost) {
+     // This makes sure we don't compile this on the device, it will kill one of
+     // the xocc hw_emu passes as it's passing a raw pointer to a kernel with no
+     // address space qualifier because we're not wrapping it in a buffer and
+     // using an accessor.
+     // We don't actually care about this parallel_for on device it's a pure
```
keryell: "since it's a pure" ?

sycl/include/CL/sycl/handler2.hpp
```diff
@@ -1117,13 +1131,19 @@ class handler {
       *PatternPtr = Pattern;
     } else {
 
+   // This makes sure we don't compile this on the device, I've not tested this
+   // one but... I'd rather not compile it on the device if it's not required
```
keryell: Turn the sentences in a more passive way

sycl/include/CL/__spir/spir_vars.hpp
agozillon: spirv_vars.hpp no longer sits inside a namespace, so this no longer sits inside a namespace, so it's on purpose in this case.

sycl/include/CL/sycl/handler2.hpp
agozillon: yep makes more sense!

agozillon:
Addressed all of the previous comments in the last commit.

Issue #25, opened 2019-05-09
[SYCL][Driver][sycl-xocc] Allow filtering of Clang arguments down to xocc compile and link stages


agozillon:
For example when a user adds -g or -O3 for debugging or optimization this should be filtered down through the xocc ToolChain to the sycl-xocc script and passed to the xocc compiler. 

There are possibly other examples of arguments that Clang and xocc share that should be passed down as well.  And perhaps allowing a user to specify certain xocc or sycl-xocc only arguments would also be useful in someway.

PR #39, opened 2019-06-11
Adding a new accessor copy related test and initial sycl-xocc "verbose" mode

agozillon:
Adding a test case for accessor to accessor D2D copy, this won't work until an XRT pull request is accepted, ill add the link in an edit once I've pull requested the relevant changes against XRT.

Also added a "verbose" mode that's currently on by default that will spit out the arguments passed to the xocc compile stage. This can be added to incrementally to make the script more debug friendly.

Relevant XRT Pull Request for Copy fix: https://github.com/Xilinx/XRT/pull/1541

### DIFF
--- a/sycl/test/xocc_tests/simple_tests/accessor_copy.cpp
+++ b/sycl/test/xocc_tests/simple_tests/accessor_copy.cpp
@@ -99,7 +99,6 @@ int main() {
       accessor<int, 2, access::mode::read, access::target::global_buffer>
           accessorFrom(idc, cgh, range<2>(Size, Size));
       cgh.single_task<noop>([=](){
-          printf("don't care force boot device \n");
       });
     });
     q.wait();


sycl/test/xocc_tests/simple_tests/accessor_copy.cpp
```diff
@@ -0,0 +1,122 @@
...
+    int Data[Size] = {0};
+    // This showcases another issue we have just now, all Kernels HAVE to have
+    // a minimum of 1 accessor or the xocc compiler will complain as there is no
+    // buffer associated with m_axi_gmem
+    buffer<int, 2> idc((int *)Data, range<2>(Size, Size));
+    q.submit([&](handler &cgh) {
+      accessor<int, 2, access::mode::read, access::target::global_buffer>
+          accessorFrom(idc, cgh, range<2>(Size, Size));
+      cgh.single_task<noop>([=](){
+          printf("don't care force boot device \n");
```
keryell: What is this `printf` doing here?

sycl/test/xocc_tests/simple_tests/accessor_copy.cpp
```diff
@@ -0,0 +1,122 @@
...
+  // things like OpenCL copy calls etc. don't work and are prone to throw
+  // runtime errors. What this does is kickstart the runtime into thinking we
+  // have an active device to map buffers to, which allows the appropriate buffer
+  // mapping to device and OpenCL calls to work.
+  {
+    const size_t Size = 20;
+    int Data[Size] = {0};
+    // This showcases another issue we have just now, all Kernels HAVE to have
+    // a minimum of 1 accessor or the xocc compiler will complain as there is no
+    // buffer associated with m_axi_gmem
```
keryell, line 96: We need to keep track of this. Create an issue if not already done?

sycl/test/xocc_tests/simple_tests/accessor_copy.cpp
```diff
@@ -0,0 +1,122 @@
...
+      assert(Data[I][J][K] == Values[I][J][K]);
+}
+
+int main() {
+  selector_defines::CompiledForDeviceSelector selector;
+  queue q{selector};
+
+  // This is a not so smart work around for avoiding cl_mem buffer errors from
+  // XRT, until we have lazy buffer generation or early program loading in place
+  // for the SYCL runtime.
```
keryell, line 83: We need to keep track of this. Create an issue if not already done?

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -37,6 +37,9 @@
 # the environment variables are a quick workaround to swap boards.
 usage() { echo xocc_compile: error: $2 >&2; exit $1; }
 
+# Make this a parameter passed in from Clang at some point
+export VERBOSE=1
```
keryell, line 41: Perhaps it is time to document the options/environment variables of the tool somewhere?

sycl/test/xocc_tests/simple_tests/accessor_copy.cpp
agozillon: I was in the process of making this: https://github.com/triSYCL/sycl/issues/40 which is tracking some of the main issues that will need a bit of thought to fix. I could create separate issues for each though.

sycl/tools/sycl-xocc/bin/sycl-xocc
agozillon: I think a lot of it should be pushed down through the compiler: https://github.com/triSYCL/sycl/issues/25 rather than be environment variables. I think VERBOSE is the only "user editable" environment variable we don't share with XRT/SDAccel e.g. XILINX_XRT and XCL_EMULATION_MODE. And VERBOSE would probably be ideal to push through the compiler arguments.

The other environment variables in that script probably shouldn't be altered/shadowed by a user that doesn't have knowledge of the script as it will probably cause problems in the compilation flow (I'm also pretty sure the script will just overwrite them at the moment). These are also documented at the top of the script.

agozillon: I am of course not against adding a .MD explaining the script in another patch though as that's helpful for everyone

keryell:
Thanks!

PR #41, opened 2019-06-14
Changing the prefix of kernels unique names from "x" to "xSYCL"

Victorlzd:
Changes to be committed:
	modified:   llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
	modified:   sycl/source/detail/program_manager/program_manager.cpp


agozillon:
Would it be possible to merge this into sycl/unified/next rather than master? Or merge this: https://github.com/triSYCL/sycl/pull/39 into next and then the next branch into master; prior to merging in these changes. I doubt it will matter too much, but the next branch is ahead of master at the moment and having all the changes on one branch would be nice rather than over 2!

keryell:
@Victorlzd I realize there is a typo in the commit message.
Can you edit `prefixe` to `prefix` and force-push again?
Thanks.

keryell:
Thanks for the contribution.

PR #46, opened 2019-07-02
[SYCL][Runtime] Minimum Modifications Required for Linux ARM compilation

agozillon:
This changes the Intel CPU intrinsics to be only included and used on 
Intel related architectures, which allows compilation for ARM. 
Unfortunately I don't have replacement calls to mimic the same 
functionallity for ARM. I don't think an equivelant exists (or at least 
I don't know enough about ARM intrinsics to work them out in a short 
space of time), it's possible at least on Linux to get this information 
from the /sys/ directory, but that's not portable and requires a chunk 
of work to create some parsing functions, I've added a related TODO.

This works for Linux, I don't think it will work for windows, but that's 
not really the goal of this change.

### DIFF
--- a/sycl/source/detail/platform_util.cpp
+++ b/sycl/source/detail/platform_util.cpp
@@ -11,7 +11,13 @@
 #include <CL/sycl/exception.hpp>
 
 #if defined(SYCL_RT_OS_LINUX)
+#if defined(__arm__) || defined(__aarch64__)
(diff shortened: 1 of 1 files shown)


keryell:
That looks like a good solution according to our minimal level of ARM processor knowledge.
Perhaps someone from ARM @kpet can provide some help?

PR #47, opened 2019-07-03
[SYCL][Runtime] Inversing the ARM Defines to fix Intel CPU

agozillon:
Made a small mistake and also pull requested before all the tests were 
complete... the #if defines were the incorrect way around for throwing 
errors. So it would throw runtime errors on intel rather than for arm! 

Sorry, silly mistake, I should be more careful..

### DIFF
--- a/sycl/source/detail/platform_util.cpp
+++ b/sycl/source/detail/platform_util.cpp
@@ -39,7 +39,7 @@ uint32_t PlatformUtil::getMaxClockFrequency() {
   throw runtime_error(
       "max_clock_frequency parameter is not supported for host device");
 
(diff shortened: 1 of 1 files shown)


keryell:
Thanks.
The reviewer should be more careful too... :-)

jeffhammond:
FYI https://github.com/intel/llvm/pull/2333 shows how to solve the cache (line) size queries with `sysconf`...

keryell:
So you mean that our special case is no longer required once merged with SYCL upstream?

jeffhammond:
Honestly, I don't know why the SYCL runtime needs to know any of that stuff, but I'm not going to worry about it right now.

PR #49, opened 2019-07-04
[SYCL][Driver][SemaSYCL][CodeGen] Add flag to switch on virtual types…

agozillon:
… (e.g. polymorphic classes) and functions

This adds a new flag that should allow virtual functions and polymorphic 
classes (although the classes could be added to a seperate flag if we 
wish more flexibility) to at least not trigger errors and for the 
compiler to emit a vtable.

Wether or not this handles everything virtual function and class 
generation requires for SYCL is another story though. It merely adds an 
option to flick some switches on in the Compiler but doesn't add any new 
AST generation for SYCL.

I think it should be fine though, so we'll see how it goes...


keryell:
Thanks.

PR #55, opened 2019-07-22
[SYCL][Tools][Compilation] Change the script sycl-xocc to handle v++ too

Victorlzd:
The script sycl-xocc now try to use v++. If it doesn't find v++, xocc will be used.

### DIFF
--- a/clang/lib/Driver/ToolChains/XOCC.cpp
+++ b/clang/lib/Driver/ToolChains/XOCC.cpp
@@ -35,10 +35,10 @@ XOCCInstallationDetector::XOCCInstallationDetector(
   // This might only work on Linux systems.
   // Rather than just checking the environment variables you could also add an
   // optional path variable for users to use.
-  auto search_program = [&] (auto programName) {
+  auto search_and_set_up_program = [&] (const char * programName) {
     llvm::ErrorOr<std::string> program = findProgramByName(programName);
     if (program) {
-      SmallString<256> vppsAbsolutePath;
+      SmallString<256> programsAbsolutePath;
       fs::real_path(*program, programsAbsolutePath);
 
       BinaryPath = programsAbsolutePath.str();
@@ -53,16 +53,14 @@ XOCCInstallationDetector::XOCCInstallationDetector(
       SDXPath = path::parent_path(programDir);
       LibPath = SDXPath + "/lnx64/lib";
 
-      // TODO: slightly stricter IsValid test.. check all strings aren't empty
+      // TODO: slightly stricter IsValid test... check all strings aren't empty
       IsValid = true;
     }
     return program;
   };
 
-  if (search_program("xocc")) {}
-  else {
-    search_program("v++");
-  }
+  if (!search_and_set_up_program("v++"))
+    search_and_set_up_program("xocc");
 }
 
 ///////////////////////////////////////////////////////////////////////////////
--- a/sycl/tools/sycl-xocc/bin/sycl-xocc
+++ b/sycl/tools/sycl-xocc/bin/sycl-xocc
@@ -35,7 +35,7 @@
 
 # TODO: Pass an argument through the driver to set options like XILINX_PLATFORM,
 # the environment variables are a quick workaround to swap boards.
-usage() { echo sycl-xocc: error: $2 >&2; exit $1; }
+usage() { echo $0: error: $2 >&2; exit $1; }
 
 # Make this a parameter passed in from Clang at some point
 export VERBOSE=1


agozillon:
Great work! I should have named these probably a little more generically originally. Although I think even sdx won't hold up if marketing have anything to do with it...

However, I think it may be worthwhile also changing the script name and the other places its required in the driver, directory and build system. From something like sycl-xocc to sycl-sdx perhaps? 

I'm also not too sure this will work in the future, if it works just now I think it works because xocc still exists somewhere in the SDx directory structure. I say that because there is an assumption in the way that the XOCC ToolChain finds the SDx directory that assumes you're looking for a binary/program called xocc: https://github.com/triSYCL/sycl/blob/sycl/unified/next/clang/lib/Driver/ToolChains/XOCC.cpp#L38

So that little function will probably need some tweaking, something like:
1) find program XOCC
2) If not found try find program V++
3) if not found, error out

But there could be a more robust solution than this.

At some point the ToolChain name/File name should probably change as well... but perhaps that's a task for the future? @keryell

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -72,14 +72,19 @@ else
...
   if [[ ! -f "$XOCC_LIB_SPIR" ]]; then
     usage 9 "could not find SDx's libspir64-39-hls.bc"
   fi
 
-  XOCC="$SDX_BIN_PATH_DIR/xocc"
-  if [[ ! -f "$XOCC" ]]; then
-    usage 10 "could not find xocc executable in $SDX_BIN_PATH_DIR"
+  SDX="$SDX_BIN_PATH_DIR/v++"
+  if [[ ! -f "$SDX" ]]; then
+    echo "[INFO] Could not find vpp executable in $SDX_BIN_PATH_DIR"
```
keryell: vpp or v++?

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -145,15 +150,15 @@ if [[ ! ${#KERNEL_NAME_ARRAY[@]} -eq 0 ]]; then
   if [[ -n "$index" ]]; then
     if [[ ! -z "$VERBOSE" ]]; then
       echo "Invoking xocc Kernel Compilation"
-      echo "xocc: $XOCC"
+      echo "xocc: $SDX"
```
keryell: So I guess it is no longer xocc even in the messages and comments everywhere...

keryell:
I am testing it...

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -144,16 +149,16 @@ if [[ ! ${#KERNEL_NAME_ARRAY[@]} -eq 0 ]]; then
   for index in "${KERNEL_NAME_ARRAY[@]}"; do
   if [[ -n "$index" ]]; then
     if [[ ! -z "$VERBOSE" ]]; then
-      echo "Invoking xocc Kernel Compilation"
-      echo "xocc: $XOCC"
+      echo "Invoking sdx Kernel Compilation"
+      echo "sdx: $SDX"
```
keryell: Perhaps these 2 lines could be replaced just by 
```
      echo "Invoking Kernel Compilation with $SDX"
```

clang/lib/Driver/ToolChains/XOCC.cpp
```diff
@@ -35,7 +35,25 @@ XOCCInstallationDetector::XOCCInstallationDetector(
...
+      BinPath = vppDir;
+
+    // TODO: Check if this assumption is correct in all installations and give
+    // environment variable specifier option or an argument to the Driver
+    SDXPath = path::parent_path(vppDir);
+    LibPath = SDXPath + "/lnx64/lib";
+
+    // TODO: slightly stricter IsValid test.. check all strings aren't empty
+    IsValid = true;
+  } else if (llvm::ErrorOr<std::string> xocc = findProgramByName("xocc")) {
```
keryell: This looks like a lot of code repetition.
You could put this in a local lambda and just call it on "`xocc"` and `"v++"`

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -1,16 +1,16 @@
...
+# SDX compiler. As SPIR/LLVM-IR is a second class citizen in SDX for the
 # moment it has some little niggling details that need worked on and the idea
 # is that this script will work around those with some aid from Clang/LLVM.
 #
-# One of the main examples is that XOCC can only compile one kernel from LLVM-BC
+# One of the main examples is that SDX can only compile one kernel from LLVM-BC
 # at a time and it requires the kernels name (also required for kernel specific
 # optimizations). This poses a problem as there can be multiple kernels in a
-# file. And when making a normal naked xocc -c command in the driver
+# file. And when making a normal naked sdx -c command in the driver
```
keryell: Use SDX instead of sdx, since sdx is not a real command anyway.

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -1,16 +1,16 @@
...
 # you won't have the neccesarry information as the command is generated before
 # the file's have even started to be compiled (Perhaps there is that I am
-# unaware of). So no kernel names and no idea how many xocc commands you'd need
+# unaware of). So no kernel names and no idea how many sdx commands you'd need
```
keryell: SDX

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -20,8 +20,8 @@
 # kernel.
 #
 # Arguments:
-# $1 == xocc's SDx bin directory where the real shell invocation that invokes
-#       the xocc compiler resides
+# $1 == sdx's SDx bin directory where the real shell invocation that invokes
```
keryell: SDX and SDX?

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -69,17 +69,22 @@ SDX_BIN_PATH_DIR="$1"
 if [[ ! -d "$SDX_BIN_PATH_DIR" ]]; then
   usage 8 "path to SDx bin directory is not a valid directory"
 else
-  # Hardcoded path from the bin directory that xocc is stored in, works for now
-  # but a future TODO might be to allow this to be optionally specified as an
-  # ENV variable
-  XOCC_LIB_SPIR="$SDX_BIN_PATH_DIR/../lnx64/lib/libspir64-39-hls.bc"
-  if [[ ! -f "$XOCC_LIB_SPIR" ]]; then
+  # Hardcoded path from the bin directory that xocc or v++ is stored in,
+  # works for nowbut a future TODO might be to allow this to be optionally
```
keryell, line 73: now but

sycl/tools/sycl-xocc/bin/sycl-xocc
keryell: By rereading all my comments again, perhaps it should be SDx everywhere?

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -35,13 +35,13 @@
...
 # the environment variables are a quick workaround to swap boards.
-usage() { echo xocc_compile: error: $2 >&2; exit $1; }
+usage() { echo sdx_compile: error: $2 >&2; exit $1; }
 
 # Make this a parameter passed in from Clang at some point
 export VERBOSE=1
 
 if [[ -z "$1" ]]; then
-  usage 1 "no sdx bin directory containing xocc"
+  usage 1 "no sdx bin directory"
```
keryell: SDx

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -35,13 +35,13 @@
 
 # TODO: Pass an argument through the driver to set options like XILINX_PLATFORM,
 # the environment variables are a quick workaround to swap boards.
-usage() { echo xocc_compile: error: $2 >&2; exit $1; }
+usage() { echo sdx_compile: error: $2 >&2; exit $1; }
```
keryell: I guess that at this level we do not know yet if it is xocc or v++ to display in the error message...

agozillon: technically you could pass another argument to the script that's basically a true false flag, if you really wanted this error to be specific. But it should probably be sycl-xocc: error: or so in reference to the script. But sdx_compile seems fine as well.

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -35,13 +35,13 @@
 
 # TODO: Pass an argument through the driver to set options like XILINX_PLATFORM,
 # the environment variables are a quick workaround to swap boards.
-usage() { echo sdx_compile: error: $2 >&2; exit $1; }
+usage() { echo sycl-xocc: error: $2 >&2; exit $1; }
```
keryell: I am thinking that replacing `sycl-xocc` by `$0` is probably the best solution, since it would just echoes the path of the script...

clang/lib/Driver/ToolChains/XOCC.cpp
```diff
@@ -35,42 +35,33 @@ XOCCInstallationDetector::XOCCInstallationDetector(
...
+      // environment variable specifier option or an argument to the Driver
+      SDXPath = path::parent_path(programDir);
+      LibPath = SDXPath + "/lnx64/lib";
 
-    // TODO: slightly stricter IsValid test.. check all strings aren't empty
-    IsValid = true;
-  } else if (llvm::ErrorOr<std::string> xocc = findProgramByName("xocc")) {
-    SmallString<256> xoccsAbsolutePath;
-    fs::real_path(*xocc, xoccsAbsolutePath);
+      // TODO: slightly stricter IsValid test.. check all strings aren't empty
```
keryell: Since you are here, replace `..` by `...`

clang/lib/Driver/ToolChains/XOCC.cpp
```diff
@@ -35,42 +35,33 @@ XOCCInstallationDetector::XOCCInstallationDetector(
...
-      BinPath = xoccDir;
-
-    // TODO: Check if this assumption is correct in all installations and give
-    // environment variable specifier option or an argument to the Driver
-    SDXPath = path::parent_path(xoccDir);
-    LibPath = SDXPath + "/lnx64/lib";
-
-    // TODO: slightly stricter IsValid test.. check all strings aren't empty
-    IsValid = true;
+  if (search_program("xocc")) {}
```
keryell: The API is a little bit convoluted and has a lot of side-effect, but this is fine and there is not need for over-engineering it more.
But renaming it and reorganizing these lines would be clearer.
Such as just:
```c++
if (!search_and_set_up_program("xocc"))
  search_and_set_up_program("v++");
```

clang/lib/Driver/ToolChains/XOCC.cpp
```diff
@@ -35,10 +35,10 @@ XOCCInstallationDetector::XOCCInstallationDetector(
   // This might only work on Linux systems.
   // Rather than just checking the environment variables you could also add an
   // optional path variable for users to use.
-  auto search_program = [&] (auto programName) {
+  auto search_and_set_up_program = [&] (const char * programName) {
```
keryell, line 38: There was an issue with `auto` in stead of `const char *` ?

keryell: OK, this is strange that the compiler is not using at least C++14...

keryell:
Thanks.

PR #59, opened 2019-07-27
[SYCL] Next Upstream Fixes

agozillon:
This should fix compilation problems with the PI interface caused by name changes for some PI buffer functions. While I was doing that I decided I'd add a new PI function to replace some old queue creation stuff wrapped by an OpenCL define, going to open a pull request in upstream SYCL for it as it's useful.

I added a flag to optionally allow variadic functions like printf in kernel code, but more importantly our SSDM intrinsics are flagged up as variadic as well. This flag is a cc1 but it's on by default for our FPGAs for now. Still requires explicit specification of the flag to turn the diagnostic error off for Intel devices/other targets.

Did the usual upping of all new SYCL test suite tests to C++17.

The biggest problem and probably most painful so far is the modification of an old existing attribute called ByVal to change the syntax to ByVal <Type>. This change unfortunately (for us) changed ByVal to require a new Attribute class, which required some Bitcode writer and reader modifications in upstream Clang/LLVM to support it. Which means we can't just alter ByVal to its old syntax style and output bitcode. So the problem is that:

1) Anything read in as bitcode with the old style ByVal attribute will be upgraded (so opt, llvm-link etc. will upgrade it)
2) Any ByVal written out, even when altered to be similar to the old style will be output in a different bitcode format than prior versions of Clang/LLVM, so it will kill xocc when it tries to read it in.
3) While you can rip it completely off and it will compile via xocc, you'll generate an invalid FPGA binary that won't execute.

The current workaround:
1) Rip all of the ByVal attributes off of function parameters
2) Re add the ByVal attributes, without the new ByVal <Type> Syntax
3) Spit out LLVM-IR from the final opt pass so it doesn't pass through the bitcode writer and enter into an unreadable state for xocc.
4) Push the altered IR through xocc's llvm-as (Clang / LLVM 7) to get our bitcode that we want to pass to xocc.

It would be possible to alter the bitcode reader/writer but I feel that would be painful long term, so I'd like to see how far the current minimalist hack can carry us. I am also of course open to other ideas!

Other than that the pass now also removes another new attribute that xocc will not read because it's not ToT. NoFree, indicates that there is no deallocation occurring in a function. Should be relatively harmless to remove after all Opt passes have been run and we no longer require it, as xocc won't be using it anyway.

------------------------------------
All tests currently pass for SDAccel 2018.3 hw_emu/sw_emu
All sw_emu tests currently pass for SDAccel 2019.1.1

The following tests have difficulty compiling for hw_emu in SDAccel 2019.1.1 at the moment:
vector_math.cpp (ICE in xocc: "Bitcode for HLS" pass)
edge_detection.cpp (ICE in xocc:  'Function Pass Manager' and 'Global Value Numbering')

The following tests compile but have run time problems for hw_emu in SDAccel 2019.1.1 at the moment:
id_mangle.cpp (XTLM Error-007:: Data Pointer)
reqd_work_group_size.cpp (XTLM Error-007:: Data Pointer)

This seems to be our System C AXI protocol library dying for some reason. 

I still can't execute single_task_vector_add in hw_emu for some reason. It compiles but I still have a simulation shell script error on my machine, seems to be the only test with this problem now after upgrading my machine. 

I'd appreciate if either of you (@keryell @Victorlzd) could run the ones with broken runtimes on your machine as well for 2019.1.1 or 2019.1, I've had a historically difficult time with hw emulation on 2019.1 on my workstation, some environment issues. 

One final open issue is that the fpga_pipes.cpp test will currently fail when running make check-all as it has ambiguity conflicts with the pipe from the system unistd.h header when there is no cl::sycl:: in-front of the pipe type to help resolve. Not too sure about this one yet.

I'll open up an issue on these tests as a reminder to look into these.

### DIFF
--- a/sycl/test/xocc_tests/issue_related/typedef_lost.cpp
+++ b/sycl/test/xocc_tests/issue_related/typedef_lost.cpp
@@ -1,0 +1,42 @@
+// RUN: true
+//
+// Existing issue where integration header appears to lose some type information
+// or more likely it just is never defined for host execution.
+// It doesn't appear to be a problem specific to the Xilinx implementation,
+// it seems like a shared problem.
+#include <cstdint>
+#include <CL/sycl.hpp>
+
+#include "../utilities/device_selectors.hpp"
+
+
+template <std::uint32_t Var>
+struct foo
+{
+    cl::sycl::accessor<int, 1, cl::sycl::access::mode::read_write> acc;
+
+    auto operator()()
+    {
+        for(auto i = 0; i < 1024; ++i)
+            acc[i] = Var;
+    }
+};
+
+auto main() -> int
+{
+    selector_defines::CompiledForDeviceSelector selector;
+
+    auto queue = cl::sycl::queue{selector};
+
+    auto buf = cl::sycl::buffer<int>{cl::sycl::range<1>{1024}};
+
+    queue.submit([&](cl::sycl::handler& cgh)
+    {
+        auto acc = buf.get_access<cl::sycl::access::mode::read_write>(cgh);
+
+        auto kernel = foo<42>{acc};
+        cgh.single_task<foo<42>>(kernel);
+    });
+
+    queue.wait();
+}


sycl/include/CL/sycl/detail/queue_impl.hpp
```diff
@@ -139,34 +139,18 @@ class queue_impl {
...
-    // TODO catch an exception and put it to list of asynchronous exceptions
-    PI_CHECK(Error);
-#endif
 
-    // Tf creating out-of-order queue failed and this property is not
+    Error = PI_CALL_RESULT(RT::piQueueCreate(Context, Device, CreationFlags,
+                                             pi::pi_cast<pi_queue *>(&Queue)));
+
+    // If creating out-of-order queue failed and this property is not
     // supported(for example, on FPGA), it will return
```
keryell: `supported (`

sycl/source/detail/pi_opencl.cpp
```diff
@@ -78,6 +78,43 @@ pi_result OCL(piextDeviceSelectBinary)(
...
+  cl_int ret_err;
+  ret_err = clGetDeviceInfo(pi_cast<cl_device_id>(device), CL_DEVICE_VERSION, 0,
+                            NULL, &devVerSize);
+
+  if (ret_err != CL_SUCCESS)
+    return pi_cast<pi_result>(ret_err);
+
+  std::string devVer(devVerSize, '\0');
+  clGetDeviceInfo(pi_cast<cl_device_id>(device), CL_DEVICE_VERSION, devVerSize,
+                  &devVer.front(), NULL);
```
keryell: `devVer.data()` instead

sycl/source/detail/pi_opencl.cpp
```diff
@@ -78,6 +78,43 @@ pi_result OCL(piextDeviceSelectBinary)(
   return PI_SUCCESS;
 }
 
+pi_result OCL(piQueueCreate)(pi_context context, pi_device device,
+                             cl_command_queue_properties properties,
+                             pi_queue *queue) {
+  size_t devVerSize;
+  cl_int ret_err;
```
keryell: Merge with next line?

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -56,8 +57,80 @@ struct XOCCIRDowngrader : public ModulePass {
...
+  /// Removes byval bitcode function parameter attribute that is applied to
+  /// pointer arguments of functions to state that they should technically be
+  /// passed by value.
+  ///
+  /// While this attribute has been long standing and has been around since
+  /// Clang/LLVM 7 it's diverged in the patch: D62319 which is the tail end of
+  /// Clang/LLVM 9.
+  ///
+  /// This LLVM patch primarily added byval <Type> syntax in the IR. And a new
+  /// Attrbiute type, which is not to be confused with anything related to the
```
keryell: Attribute

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -56,8 +57,80 @@ struct XOCCIRDowngrader : public ModulePass {
...
+  /// type system. It basically just states the attribute is carrying around
+  /// some Type/Value information rather than string/int/enum and gives it
+  /// somewhere to store it.
+  ///
+  /// However Now the bitcode writer will write out some different bitcode for
```
keryell: now

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -56,8 +57,80 @@ struct XOCCIRDowngrader : public ModulePass {
...
+  /// byval that is unreadable by xocc and the bitcode reader upgrades any byval
+  /// attributes to the new byval <Type> syntax. The way we currently work
+  /// around this is:
+  /// 1) Delete the ByVal attribute, erasing any byval <Type> syntax in the IR
+  /// and then re-adding the old byval with no <Type> Syntax.
+  /// 2) In the SYCL xocc script when we run this pass with opt we make sure we
+  /// emit LLVM IR and not bitcode after this pass so that it doesn't run through
```
keryell: LLVM IR and not bitcode: unclear since the bitcode is LLVM IR

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -73,6 +73,8 @@ else
   # but a future TODO might be to allow this to be optionally specified as an
   # ENV variable
   XOCC_LIB_SPIR="$SDX_BIN_PATH_DIR/../lnx64/lib/libspir64-39-hls.bc"
+  # It says 3.9, but trust us this is Clang/LLVM 7..
```
keryell, line 77: The problem is that if some day it is no longer version 7, we can no longer trust this comment...
So let's try to mitigate this with something like:
```bash
# It says 3.9, but this is Clang/LLVM 7 according to some urban legend...
```

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -127,7 +129,20 @@ $LLVM_LINK "$TEMPORARY_DIR/${SOURCE_FILE_NAME}_kernels-optimized.bc" \
 # Using the LLVM Linker will add more incompatible IR so run the downgrader in
 # a second pass afterwards, rather than once before in the first opt pass
 # and then again after linking.
-$OPT -O3 -xoccIRDowngrader \
+#
+# We also only wish to emit LLVM IR and not bitcode as we currently use the
```
keryell: `LLVM IR assembly language format`

sycl/source/detail/pi_opencl.cpp
agozillon: going to leave it as .front() for now as the Upstream implementation will probably need to remain .front() and changing this will just cause some pull request conflicts.

sycl/include/CL/sycl/detail/queue_impl.hpp
agozillon: changing this in upstream, so next synch will fix it.

agozillon:
I'd hold off merging this until the Pi.h modification pull request goes into the Upstream Intel implementation and I can replicate the changes here. Happy to address additional review comments in the mean time!

sycl/source/detail/pi_opencl.cpp
keryell: Actually you are right, it does not work before C++17. Since `devVer.front()` is a reference according to the C++ specification it works and is even shorten than `&*devVer.begin()`

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -62,20 +62,20 @@ struct XOCCIRDowngrader : public ModulePass {
   /// Clang/LLVM 9.
   ///
   /// This LLVM patch primarily added byval <Type> syntax in the IR. And a new
-  /// Attrbiute type, which is not to be confused with anything related to the
+  /// attrbiute type, which is not to be confused with anything related to the
```
keryell: I am unsure it is a great progress... ;-)

agozillon: didn't realize it was misspelled as well, sorry. Which is irritating as I had my spell checker on...

agozillon: should be fixed now

agozillon:
rebased on recent changes because of merge conflict and force pushed.

agozillon:
@keryell can be merged whenever you get a spare minute

PR #71, opened 2019-10-04
Fix to Next Branch 

agozillon:
This fixes a few leftover things in the runtime (that I missed when I originally merged in and pushed directly to next). 

Adds some file cleanup to the script.

Reworked it to use the infer-address-space pass as suggested by the people over at Intel (thanks). So hardware emulation is a little more stable, but a long way away from perfect still. 

Some changes to passes to cleanup the IR from upgrading to further upstream LLVM and a small tweak to a pass to deal with the address space casts left over around the SSDM intrinsics.

Moved some tests around that are now fixed (on merge a couple of issues can likely be closed), some still remain broken sadly. But overall things seem a little more consistent, in at least 2019.1.

Added some new tests, one in particular will be a painful fix which is local_memory.cpp, local memory is not being correctly inferred/defined for our FPGAs.

I still need to test these run properly in hw_emulation (need to compile and execute on a separate machine) which ill do over the next day or two, but that shouldn't prevent the merge/refactor of this patch.

### DIFF
--- a/llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
+++ b/llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
@@ -25,13 +25,13 @@
 #include "llvm/Demangle/Demangle.h"
 #include "llvm/IR/Argument.h"
 #include "llvm/IR/CallingConv.h"
+#include "llvm/IR/CallSite.h"
 #include "llvm/IR/Constants.h"
 #include "llvm/IR/Function.h"
 #include "llvm/IR/InstIterator.h"
 #include "llvm/IR/Metadata.h"
 #include "llvm/IR/Module.h"
 #include "llvm/IR/Type.h"
-#include "llvm/IR/CallSite.h"
 #include "llvm/Pass.h"
 #include "llvm/Support/Debug.h"
 #include "llvm/Support/raw_ostream.h"
@@ -380,10 +380,10 @@ struct InSPIRation : public ModulePass {
   /// leakage however, but should result in more overall address space
   /// consistency/stability due to the addition of more concrete address spaces.
   void handleSpecArrayPartition(CallInst *CI) {
-    for (unsigned int i = 0; i < CI->getNumOperands(); ++i) {
-      if (CI->getOperand(i)->getType()->isPointerTy()) {
-        if (auto *ASC = dyn_cast<AddrSpaceCastInst>(CI->getOperand(i))) {
-          if (ASC->getDestAddressSpace() == /*Generic Address Space*/4) {
+    for (auto &Op : CI->operands()) {
+      if (Op->getType()->isPointerTy()) {
+        if (auto *ASC = dyn_cast<AddrSpaceCastInst>(Op)) {
+          if (ASC->getDestAddressSpace() == /*Generic AS*/ 4) {
             ASC->replaceAllUsesWith(ASC->getPointerOperand());
             ASC->eraseFromParent();
           }
@@ -404,8 +404,8 @@ struct InSPIRation : public ModulePass {
   /// InferAddressSpaces pass and teach it to deal with these SSDM calls as
   /// Intrinsics.
   void ssdmAddressSpaceFix(Function &F) {
-    for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I)
-      if (auto *Call = dyn_cast<CallInst>(&*I))
+    for (auto &I : instructions(F))
+      if (auto *Call = dyn_cast<CallInst>(&I))
         if (auto Func = dyn_cast<Function>(Call->getCalledFunction()))
           if (Func->getName() == "_ssdm_SpecArrayPartition")
             handleSpecArrayPartition(Call);
@@ -451,7 +451,7 @@ struct InSPIRation : public ModulePass {
         /// function to be sycl_func_x, if xocc ever gets a little friendlier to
         /// spir input, probably not required.
         } else if (isTransitiveNonIntrinsicFunc(F)
-                    && !F.isDeclaration()) {
+                   && !F.isDeclaration()) {
         // After kernels code selection, there are only two kinds of functions
         // left: funcions called by kernels or LLVM intrinsic functions.
         // For functions called in SYCL kernels, put SPIR calling convention.
@@ -475,11 +475,11 @@ struct InSPIRation : public ModulePass {
         // related to HLS needing names to generate XML).
         //
         // It doesn't require application to the SPIR intrinsics as we're
-        // linking against the HLS SPIR library, which is already conform-ant.
+        // linking against the HLS SPIR library, which is already conformant.
         giveNameToArguments(F);
         ssdmAddressSpaceFix(F);
       } else if (isTransitiveNonIntrinsicFunc(F)
-                  && F.isDeclaration()) {
+                 && F.isDeclaration()) {
         // push back intrinsics to make sure we handle naming after changing the
         // name of all functions to sycl_func.
         // Note: if we do not rename all the functions to sycl_func_N, a more
--- a/llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
+++ b/llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
@@ -19,6 +19,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/IR/Attributes.h"
 #include "llvm/IR/Function.h"
+#include "llvm/IR/InstIterator.h"
 #include "llvm/IR/InstrTypes.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Pass.h"
@@ -131,18 +132,15 @@ struct XOCCIRDowngrader : public ModulePass {
       }
 
       // These appear on Call/Invoke Instructions as well
-      for (auto &BB : F) {
-        for (auto &I : BB) {
-          if (CallBase *CB = dyn_cast<CallBase>(&I)) {
-            for (unsigned int i = 0; i < CB->getNumArgOperands(); ++i) {
-              if (CB->paramHasAttr(i, llvm::Attribute::ByVal)) {
-                CB->removeParamAttr(i, llvm::Attribute::ByVal);
-                CB->addParamAttr(i, llvm::Attribute::ByVal);
-              }
+      for (auto &I : instructions(F))
+        if (CallBase *CB = dyn_cast<CallBase>(&I)) {
+          for (unsigned int i = 0; i < CB->getNumArgOperands(); ++i) {
+            if (CB->paramHasAttr(i, llvm::Attribute::ByVal)) {
+              CB->removeParamAttr(i, llvm::Attribute::ByVal);
+              CB->addParamAttr(i, llvm::Attribute::ByVal);
             }
           }
         }
-      }
     }
   }
 
--- a/sycl/test/xocc_tests/issue_related/issue_70.cpp
+++ b/sycl/test/xocc_tests/issue_related/issue_70.cpp
@@ -53,17 +53,12 @@ auto main() -> int
     auto s_buf = cl::sycl::buffer<std::size_t, 3>{
                     cl::sycl::range<3>{42, 42, 42}};
 
-    q.submit([&](cl::sycl::handler& cgh)
-    {
+    q.submit([&](cl::sycl::handler& cgh) {
         auto s = s_buf.get_access<cl::sycl::access::mode::read_write>(cgh);
 
-        cgh.single_task<burst_test>([=]()
-        {
-            printf("I enter Kernel \n");
+        cgh.single_task<burst_test>([=]() {
             loop(s);
-            printf("I exit Kernel \n");
         });
-
     });
     q.wait();
 
--- a/sycl/test/xocc_tests/issue_related/local_memory.cpp
+++ b/sycl/test/xocc_tests/issue_related/local_memory.cpp
@@ -49,9 +49,6 @@ int main() {
 
    cgh.parallel_for<simple_local>(nd_range<1>(range<1>(sz), range<1>(sz / 2)),
       [=](nd_item<1> item) {
-        printf("get_local_id: %d \n", item.get_local_id()[0]);
-        printf("get_global_id: %d \n", item.get_global_id()[0]);
-
         local_mem[item.get_local_id()[0]] += item.get_local_id()[0];
         acc_g[item.get_global_id()[0]] = local_mem[item.get_local_id()[0]];
       });


llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -369,21 +452,32 @@ struct InSPIRation : public ModulePass {
...
+        F.setName("sycl_func_" + Twine{funcCount++});
+
+        // While functions do come "named" it's in the form %0, %1 and XOCC
+        // doesn't like this for the moment. XOCC demands function arguments
+        // be either unnamed or named non-numerically. This is a separate
+        // issue from the reason we name kernel arguments (which is more
+        // related to HLS needing names to generate XML).
+        //
+        // It doesn't require application to the SPIR intrinsics as we're
+        // linking against the HLS SPIR library, which is already conform-ant.
```
keryell: conformant?

sycl/include/CL/__spir/spir_vars.hpp
```diff
@@ -19,13 +19,13 @@
...
 
 // These are converted to SPIR builtins by the inSPIRation pass.
-size_t __spir_ocl_get_global_size(uint dimindx);
-size_t __spir_ocl_get_local_size(uint dimindx);
-size_t __spir_ocl_get_global_id(uint dimindx);
-size_t __spir_ocl_get_local_id(uint dimindx);
-size_t __spir_ocl_get_global_offset(uint dimindx);
-size_t __spir_ocl_get_group_id(uint dimindx);
-size_t __spir_ocl_get_num_groups(uint dimindx);
+size_t __spir_ocl_get_global_size(unsigned int dimindx);
```
keryell, line 22: I am curious about where these `uint` were coming from at the first place...

sycl/test/xocc_tests/issue_related/issue_70.cpp
```diff
@@ -0,0 +1,71 @@
...
+    auto s_buf = cl::sycl::buffer<std::size_t, 3>{
+                    cl::sycl::range<3>{42, 42, 42}};
+
+    q.submit([&](cl::sycl::handler& cgh)
+    {
+        auto s = s_buf.get_access<cl::sycl::access::mode::read_write>(cgh);
+
+        cgh.single_task<burst_test>([=]()
+        {
+            printf("I enter Kernel \n");
```
keryell: I am suffering every time I look at `printf`, even more when it should be a `puts`... :-)

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -337,6 +338,79 @@ struct InSPIRation : public ModulePass {
...
+  /// this will result in left over generic casts. This function is here
+  /// to deal with the cases of the left over generics caused by SSDM intrinsics
+  /// as if they're left in the compilation will fail.
+  ///
+  /// In the future if we ever define an LLVM target backend similar to AMDGPU
+  /// and we end up with a lot of these edge cases we could move this to the
+  /// InferAddressSpaces pass and teach it to deal with these SSDM calls as
+  /// Intrinsics.
+  void ssdmAddressSpaceFix(Function &F) {
+    for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I)
```
keryell:
```suggestion
    for (auto *i : instructions(F))
```

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -337,6 +338,79 @@ struct InSPIRation : public ModulePass {
...
+      if (auto *Call = dyn_cast<CallInst>(&*I))
```
keryell:
```suggestion
      if (auto *Call = dyn_cast<CallInst>(I))
```

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -337,6 +338,79 @@ struct InSPIRation : public ModulePass {
...
+  ///
+  /// Note: Unsure how robust this is with the small sample size we currently
+  /// have to test against. So it's a WIP. If the situation becomes untenable
+  /// we can likely revert the accessor class back to it's prior form by
+  /// reverting this pull: https://github.com/intel/llvm/pull/348 (commit:
+  /// 609999c4e1aeca05aff010ce5e2eb08dde08fd69). This may cause address space
+  /// leakage however, but should result in more overall address space
+  /// consistency/stability due to the addition of more concrete address spaces.
+  void handleSpecArrayPartition(CallInst *CI) {
+    for (unsigned int i = 0; i < CI->getNumOperands(); ++i) {
```
keryell:
```suggestion
    for (auto *Op : operands(CI)) {
```

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -337,6 +338,79 @@ struct InSPIRation : public ModulePass {
...
+      if (CI->getOperand(i)->getType()->isPointerTy()) {
```
keryell:
```suggestion
      if (Op->getType()->isPointerTy()) {
```

sycl/test/xocc_tests/issue_related/local_memory.cpp
```diff
@@ -0,0 +1,69 @@
...
+  // This test only really funcitonally shows that it's not using global memory
+  // in place of local memory...but it could also be implemented via private..
+  q.submit([&](handler &cgh) {
+    auto acc_g = a.get_access<access::mode::write>(cgh);
+    accessor<int, 1, access::mode::read_write, access::target::local>
+        local_mem(range<1>(sz / 2), cgh);
+
+   cgh.parallel_for<simple_local>(nd_range<1>(range<1>(sz), range<1>(sz / 2)),
+      [=](nd_item<1> item) {
+        printf("get_local_id: %d \n", item.get_local_id()[0]);
```
keryell: Oh, some `printf` with a lovely `#include <iostream>`! :-)

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -369,21 +452,32 @@ struct InSPIRation : public ModulePass {
         /// spir input, probably not required.
         } else if (isTransitiveNonIntrinsicFunc(F)
                     && !F.isDeclaration()) {
```
keryell: Indentation issue?

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -102,10 +124,24 @@ struct XOCCIRDowngrader : public ModulePass {
...
-              P.addAttr(llvm::Attribute::ByVal);
+         if (P.hasAttribute(llvm::Attribute::ByVal)) {
+             P.removeAttr(llvm::Attribute::ByVal);
+             P.addAttr(llvm::Attribute::ByVal);
+         }
+      }
+
+      // These appear on Call/Invoke Instructions as well
+      for (auto &BB : F) {
+        for (auto &I : BB) {
```
keryell: See above to use directly a function `instructions` range instead

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -337,6 +338,79 @@ struct InSPIRation : public ModulePass {
...
+  /// we can likely revert the accessor class back to it's prior form by
+  /// reverting this pull: https://github.com/intel/llvm/pull/348 (commit:
+  /// 609999c4e1aeca05aff010ce5e2eb08dde08fd69). This may cause address space
+  /// leakage however, but should result in more overall address space
+  /// consistency/stability due to the addition of more concrete address spaces.
+  void handleSpecArrayPartition(CallInst *CI) {
+    for (unsigned int i = 0; i < CI->getNumOperands(); ++i) {
+      if (CI->getOperand(i)->getType()->isPointerTy()) {
+        if (auto *ASC = dyn_cast<AddrSpaceCastInst>(CI->getOperand(i))) {
+          if (ASC->getDestAddressSpace() == /*Generic Address Space*/4) {
```
keryell: Should it be the argument passed to `-flat-address-space` instead?

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -102,10 +124,24 @@ struct XOCCIRDowngrader : public ModulePass {
...
+          if (CallBase *CB = dyn_cast<CallBase>(&I)) {
+            for (unsigned int i = 0; i < CB->getNumArgOperands(); ++i) {
+              if (CB->paramHasAttr(i, llvm::Attribute::ByVal)) {
+                CB->removeParamAttr(i, llvm::Attribute::ByVal);
+                CB->addParamAttr(i, llvm::Attribute::ByVal);
```
keryell: Why do you need to remove it and to add it again if it does already exist?

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
```diff
@@ -102,10 +124,24 @@ struct XOCCIRDowngrader : public ModulePass {
...
+             P.addAttr(llvm::Attribute::ByVal);
+         }
+      }
+
+      // These appear on Call/Invoke Instructions as well
+      for (auto &BB : F) {
+        for (auto &I : BB) {
+          if (CallBase *CB = dyn_cast<CallBase>(&I)) {
+            for (unsigned int i = 0; i < CB->getNumArgOperands(); ++i) {
+              if (CB->paramHasAttr(i, llvm::Attribute::ByVal)) {
```
keryell: It is less clear how to simplify this...
Perhaps starting with
```
            for (auto *Op : CB->args()) {
```
but I do not know how to go on with a range interface...
It looks there the attribute interface is very oriented towards operand index...

sycl/doc/GettingStartedAlveo.md
```diff
@@ -294,9 +294,6 @@ export XILINX_XRT=/opt/xilinx/xrt
 SYCL_HOME=/var/tmp/rkeryell/SYCL/sycl
 SYCL_BIN_DIR=$SYCL_HOME/build/bin
 export LD_LIBRARY_PATH=$XILINX_XRT/lib:$SYCL_HOME/build/lib:$LD_LIBRARY_PATH
-# Workaround a current address space issues in the SYCL compiler
-# https://github.com/triSYCL/sycl/pull/51
-export DISABLE_INFER_AS=1
```
keryell, line 299: I love when a work-around goes away! \o/

sycl/test/xocc_tests/issue_related/issue_70.cpp
agozillon: woops, forgot to remove that...

sycl/test/xocc_tests/issue_related/local_memory.cpp
agozillon: OpenCL has printf! I'm not so sure about iostream! but I don't think I meant to leave this in either...

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
agozillon: That argument goes to a different pass, but I could add it as part of InSPIRation as well and see if it will pass it into here, so it will stay synchronized.

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
agozillon: ByVal tracks type information in Clang/LLVM 9+, which will break anything prior to the patch that added it as it isn't backwards compatible. It existed before Clang/LLVM 9 in a different format (no tracking of type information) so you cannot completely remove the Attribute as it holds relevant information and will break the xocc backend if you remove it. 

All this does is reset it to it's default state which is what it looked like prior to the Clang/LLVM change.

sycl/test/xocc_tests/issue_related/local_memory.cpp
keryell: It was a joke anyway.
By the way, I think there is nothing in the SYCL standard that says that `printf` is expected to work, even if you have an OpenCL back-end...

sycl/include/CL/__spir/spir_vars.hpp
agozillon: not too sure I can answer that actually, a change upstream made it break, so perhaps it was reordering and/or removal of includes

sycl/test/xocc_tests/issue_related/local_memory.cpp
agozillon: I'm pretty sure you are correct there isn't anything stating it's required to work/isn't required to work, it's implementation defined I believe; I remember a discussion a few months ago on it I think. 

It isn't implemented for Intel anyway, it will flag up as a diagnostic error as printf is a variadic function. So it really shouldn't be used inside of a kernel, I left it in this case by accident as I use them for quick testing on our FPGAs as they support printf reasonably well in emulation for now!

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
agozillon: https://github.com/triSYCL/sycl/blob/5e48bc5a396ae44d2eb4eb8e7fa44b1526b73c9e/llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp#L78 

In reference to our conversation, a link to the comment on it! Read over it and see if you'd like me to add anything else in case it is not fully covered.

llvm/lib/SYCL/XOCCIRDowngrader/XOCCIRDowngrader.cpp
agozillon: Yes, there's a few hold over APIs that don't really go well with ranges, I'll have a look and see if I can find something a little more appealing

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
agozillon: Seems like you cannot register the same named argument for multiple passes and the method to share pass options across pass boundaries seems a little intrusive (the InferAddressSpace pass also doesn't have a header which makes the change a little weirder): http://llvm.org/docs/CommandLine.html#internal-vs-external-storage

This seems like it would be the ideal thing to use, but it doesn't seem to have any guarantees that the options will be registered when we need them and the type it returns doesn't really seem to allow access to the value in any case:
http://llvm.org/docs/CommandLine.html#the-cl-getregisteredoptions-function

I could register another argument with a different name for this pass that will accept an identical value or we could leave it as-is for now?

llvm/lib/SYCL/InSPIRation/InSPIRation.cpp
```diff
@@ -404,8 +404,8 @@ struct InSPIRation : public ModulePass {
   /// InferAddressSpaces pass and teach it to deal with these SSDM calls as
   /// Intrinsics.
   void ssdmAddressSpaceFix(Function &F) {
-    for (inst_iterator I = inst_begin(F), E = inst_end(F); I != E; ++I)
-      if (auto *Call = dyn_cast<CallInst>(&*I))
+    for (auto &I : instructions(F))
```
keryell, line 407: It is not possible to get directly the pointer with `auto *I`?

agozillon: It has some trouble deducing to auto *, and it's used with auto & in the other locations throughout LLVM that the instructions call is made. I've found that there's a couple of the Clang APIs that won't deduce auto *'s.

keryell:
Thanks.

PR #75, opened 2020-03-24
[NFC] Fix build with shared libraries

Ralender:
Some LLVMBuild.txt where not kept up to date so the shared library build was failing. this patch fixes this.


keryell:
Thanks for improving the build system.

PR #77, opened 2020-09-04
Fix const_iterator in FPGA partition_array

keryell:
This fix a compilation error where the const iterator were not actually const.


Ralender:
This seems fine.

PR #87, opened 2020-09-30
[SYCL] Fix edge_detection and add it to the lit test-suite

Ralender:
edge_detection was failing because of the first argument was expected of xilinx_partition_array was in the generic address space instead of the default/stack address space.
I also added edge_detection  to the test-suite it will run only if opencv4 is found and will only compile in hw_emu because images are too large datasets.


sycl/test/xocc_tests/edge_detection/edge_detection.cpp
```diff
@@ -92,10 +101,10 @@ int main(int argc, char* argv[]) {
     // to parallel_fors with local sizes
      [=]() {
       auto gX = xilinx::partition_array<char, 9,
-                xilinx::partition::complete<0>>({-1, 0, 1, -2, 0, 2, -1, 0, 1});
+                xilinx::partition::complete<1>>({-1, 0, 1, -2, 0, 2, -1, 0, 1});
```
keryell, line 104: Curious this change.
Are you changing the API?
Do you think it is better to start at 1?

sycl/test/lit.cfg.py
```diff
@@ -258,13 +266,25 @@ def getDeviceCount(device_type):
     if xocc == "only":
         config.name = 'SYCL-XOCC'
         config.test_source_root = config.test_source_root + "/xocc_tests"
+    run_if_hw="echo"
```
keryell, line 269: At least add some motivating comments

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
```diff
@@ -59,7 +53,7 @@ void dataflow(T functor) {
     pipeline way.
 */
 template <typename T>
-SYCL_DEVICE_ANNOTATE("xilinx_pipeline") __attribute__((always_inline))
+__SYCL_DEVICE_ANNOTATE("xilinx_pipeline") __attribute__((always_inline))
```
keryell, line 56: Could even be `__TRISYCL` prefix to reduce conflicts with Intel or other SYCL implementations

llvm/lib/SYCL/InSPIRation.cpp
```diff
@@ -348,78 +348,6 @@ struct InSPIRation : public ModulePass {
     }
   }
 
-  /// This currently looks through the arguments passed to the SSDM intrinsic
```
keryell, line 351: So you are removing the support for the old style SSDM?
I guess it is no longer used anyway @yu810226 @aisoard 
By the way, I am expecting the same level of comments elsewhere in the code... :-)

keryell (commented):
I have the feeling that now array partitioning is no longer used by the SYCL backend.
Note that it does not really change since it was ignored by HLS before... :-)

PR #88, opened 2020-10-02
Remove the need for using XCL_EMULATION_MODE

Ralender:


clang/lib/Driver/Driver.cpp
```diff
@@ -674,6 +674,25 @@ static bool isValidSYCLTriple(llvm::Triple T) {
   return true;
 }
 
+static llvm::Triple completeSYCLTriple(llvm::Triple T) {
```
keryell, line 678: Probably you can describe what the function does at high-level and intersperse a few comments about the logic.
I am unsure that everybody knows the world-famous `XCL_EMULATION_MODE` environment variable...

clang/lib/Driver/ToolChains/XOCC.cpp
```diff
@@ -185,10 +183,28 @@ void SYCL::LinkerXOCC::constructSYCLXOCCCommand(
...
   AddForwardedOptions(Args, CmdArgs, options::OPT_Xsycl_backend,
-      options::OPT_Xsycl_backend_EQ, getTargetTriple(), C.getDriver());
+      options::OPT_Xsycl_backend_EQ, TC.getTriple(), C.getDriver());
+
+  // Script Arg $8
   AddForwardedOptions(Args, CmdArgs, options::OPT_Xsycl_linker,
-      options::OPT_Xsycl_linker_EQ, getTargetTriple(), C.getDriver());
+      options::OPT_Xsycl_linker_EQ, TC.getTriple(), C.getDriver());
+
+  switch (TC.getTriple().getSubArch()) {
```
keryell, line 194: Add a few comments

llvm/include/llvm/ADT/Triple.h
```diff
@@ -102,8 +102,8 @@ class Triple {
     fpga_aocr,      // Intel FPGA: linked early image
     fpga_aocx,      // Intel FPGA: linked image
     fpga_dep,       // Intel FPGA: dependency file
-    fpga32,         // 32-bit FPGA
-    fpga64,         // 64-bit FPGA
+    fpga32,         // 32-bit Xilinx FPGA
```
keryell, line 105: Are there any conflicts with Intel FPGA?
I assume we can have both Xilinx and Intel FPGA on the same machine.

llvm/lib/SYCL/ASFixer/LLVMBuild.txt
```diff
@@ -1,21 +0,0 @@
-;===- ./lib/SYCL/LLVMBuild.txt ------------------------------*- Conf -*-----===;
```
keryell, line 1: I added them from a merge with `sycl/unified/master` to avoid loosing https://github.com/triSYCL/sycl/pull/75
But I have no idea whether they are required or not anymore.

sycl/doc/GettingStartedXilinxFPGA.md
```diff
@@ -355,18 +355,27 @@ To run an example from the provided examples:
...
-  # Do not forget about unset the variable for real hardware execution
-  unset XCL_EMULATION_MODE
   # Run on the real FPGA board
   ./parallel_for_ND_range
   ```
-
 Note that in the previous examples, the compilation line does not
 change, only the environment ``XCL_EMULATION_MODE`` variable.
 
+it is possible to not use XCL_EMULATION_MODE and specify only the beginning of the triple:
```
keryell:
```suggestion
it is possible to not use `XCL_EMULATION_MODE` and specify only the beginning of the triple:
```

sycl/plugins/opencl/pi_opencl.cpp
```diff
@@ -257,7 +257,8 @@ pi_result piextDeviceSelectBinary(pi_device device, pi_device_binary *images,
   // Find the appropriate device image, fallback to spirv if not found
   pi_uint32 fallback = invalid_ind;
   for (pi_uint32 i = 0; i < num_images; ++i) {
-    if (strcmp(images[i]->DeviceTargetSpec, image_target) == 0) {
+    if (strncmp(images[i]->DeviceTargetSpec, image_target,
+                strlen(image_target)) == 0) {
```
keryell, line 262: Add a comment about why you need this change.
I do not understand why.
I guess we can remove line 248.

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,28 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
   return Result;
 }
 
+static void handleXilinxKernels(RTDeviceBinaryImage* Img) {
+  static bool has_been_invoked = false;
+  const char* target = Img->getTarget();
+  const char* env_var = "XCL_EMULATION_MODE";
```
keryell:
```suggestion
  auto constexpr env_var = "XCL_EMULATION_MODE";
```

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,28 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
   return Result;
 }
 
+static void handleXilinxKernels(RTDeviceBinaryImage* Img) {
```
keryell: Add a high-level comment and a few comments inside the code

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,28 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
...
+  static bool has_been_invoked = false;
+  const char* target = Img->getTarget();
+  const char* env_var = "XCL_EMULATION_MODE";
+  if (strncmp("fpga64_", target, 7) == 0 || strncmp("fpga32", target, 7) == 0) {
+    target += 7;
+    if (strcmp(target, "hw") == 0) {
+      if (has_been_invoked)
+        assert(!std::getenv(env_var));
```
keryell: I have the feeling that all these asserts should actually be replaced by some SYCL exception `throw`.
Because the runtime could be compiled with `-DNDEBUG` to remove the `assert()` macros leaving a lot of undefined behavior all over the place.

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,28 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
...
+      ::unsetenv(env_var);
+    }
+    else {
+      if (has_been_invoked) {
+        assert(std::getenv(env_var));
+        assert(strcmp(std::getenv(env_var), target) ==0);
+      }
+      ::setenv(env_var, target, true);
```
keryell, line 921: I guess it is useless to reset the value if you checked it before?
Or perhaps all this `has_been_invoked` optimization does not worth the effort?

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,28 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
   return Result;
 }
 
+static void handleXilinxKernels(RTDeviceBinaryImage* Img) {
+  static bool has_been_invoked = false;
+  const char* target = Img->getTarget();
+  const char* env_var = "XCL_EMULATION_MODE";
+  if (strncmp("fpga64_", target, 7) == 0 || strncmp("fpga32", target, 7) == 0) {
```
keryell:
```suggestion
  if (strncmp("fpga64_", target, 7) == 0 || strncmp("fpga32_", target, 7) == 0) {
```

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -937,6 +959,7 @@ void ProgramManager::addImages(pi_device_binaries DeviceBinary) {
     if (KSId == 0)
       KSId = getNextKernelSetId();
 
+    handleXilinxKernels(Img.get());
```
keryell: The name is a little bit misleading since it will actually handle other kernels too (by doing nothing...)

sycl/test/lit.cfg.py
```diff
@@ -207,6 +210,10 @@ def getDeviceCount(device_type):
     if "XCL_EMULATION_MODE" in os.environ:
         if os.environ["XCL_EMULATION_MODE"] == "hw":
             acc_run_substitute="env -u XCL_EMULATION_MODE " + acc_run_substitute
+    if xocc_target != "hw_emu":
```
keryell, line 220: Add a few comments in the neighborhood.
I do not use `setsid` and `flock` for breakfast...
If you use options, use explicit ones, like `--exclusive` instead of `-x`.

sycl/tools/sycl-xocc/bin/sycl-xocc
```diff
@@ -32,6 +32,9 @@
 #       files
 # $6 == The name we wish to give our final output .xcl binary, important that it
 #       is consistent with what the driver expects it to be.
+# $7 == The name of the file containing addtional args for v++ -c
```
keryell:
```suggestion
# $7 == The name of the file containing additional args for v++ -c
```
Use a spell-checker.
And fix the same error everywhere in this file.

keryell (changes requested):
That looks like a great improvement in usability!
Just a few remarks...

sycl/source/detail/program_manager/program_manager.cpp
Ralender: we could throw be this code is run duding startup of the executable before main so the user as no chance of catching that throw.

sycl/source/detail/program_manager/program_manager.cpp
Ralender: it is not an optimization it is a correctness check. this patch doesn't support mixing sw_emu with hw kernels in the same executable.

llvm/include/llvm/ADT/Triple.h
Ralender: > Are there any conflicts with Intel FPGA?

not that I am aware of
> I assume we can have both Xilinx and Intel FPGA on the same machine.

I can't think of any reason it wouldn't be possible but I don't expect anyone has tried and I wouldn't be surprised if there was issue doing it. we have no tests where not all kernels are targeting the same device. 

the comment gave the impression that this target is generic when it isn't.
maybe it should be renamed.
also I am not sure there is any value to having fpga32 since to my understanding fpga don't have fixed bit width like traditional hardware.

llvm/lib/SYCL/ASFixer/LLVMBuild.txt
Ralender: the file structure was recently flattened to be more inline with the rest of LLVM and to make it easier to add new passes.
so these files aren't useful anymore because there directory is empty.

llvm/include/llvm/ADT/Triple.h
agozillon: The target was named/commented generically because Intel hadn't really added their FPGA target yet as I remember, so I had some naive hope that it might lineup! Of course, I was wrong :-)

As for the 32/64 bit, it's mainly as the SPIR target we use has 32/64 bit cases  (https://github.com/triSYCL/sycl/blob/98a61101707652d38e71ebb8938e035f745133da/clang/lib/Basic/Targets.cpp#L556) and I believe this has some minor affect on the LinuxTargetInfo/OSTargetInfo as well (this is likely negligible for SPIR though). So it simplified adding it. I am no expert in the TargetInfo system, so there is without a doubt a better solution :-D

In essence feel free to rename them whatever you want and/or remove one! Perhaps the HLS team have a good naming convention.

keryell: @aisoard @yu810226 for bike-shedding...
Otherwise, can we have the vendor in another place of the triple?

sycl/source/detail/program_manager/program_manager.cpp
keryell: I do not really see the logic at this level. So you are talking about the case you have several binary images, each one is handled by a different call to this function and you use the `XCL_EMULATION_MODE` environment variable as a global variable to pass checking information between each called instance?
But even in this case I have the feeling that there is a missing return after lines 902 and 919.

sycl/test/lit.cfg.py
```diff
@@ -200,13 +203,24 @@ def getDeviceCount(device_type):
     lit_config.warning("Accelerator device not found")
 
 if xocc != "off":
+    # xrt doesn't deal well with multiple executable using it concurrently.
+    # so we wrap every use of XRT inside an file lock
+    # we also wrap invocation of executable in an setsid to prevent
+    # a single program faillure from ending all the tests.
```
keryell:
```suggestion
    # XRT doesn't deal well with multiple executable using it concurrently.
    # So we wrap every use of XRT inside an file lock.
    # We also wrap invocation of executable in an setsid to prevent
    # a single program faillure from ending all the tests.
```
Thanks for the very helpful comments!
@stsoe: is this a bug in our SYCL runtime about not using correctly XRT (because for example we should specify some exclusive device use) or XRT does not allow this use case?

sycl/test/lit.cfg.py
```diff
@@ -200,13 +203,24 @@ def getDeviceCount(device_type):
...
     xrt_lock="/tmp/xrt.lock"
-    acc_run_substitute+= "setsid flock -x " + xrt_lock + " "
+    acc_run_substitute+= "setsid flock --exclusive " + xrt_lock + " "
     if os.path.exists(xrt_lock):
         os.remove(xrt_lock)
+    # XCL_EMULATION_MODE = hw is only valid for our sycl driver,
```
keryell:
```suggestion
    # XCL_EMULATION_MODE = hw is only valid for our SYCL driver, 
```

keryell (changes requested):
Thanks for the clarifications.
I still have a few questions...

sycl/source/detail/program_manager/program_manager.cpp
keryell: Ah good point.
Add this into a comment.
Otherwise, perhaps we should have the runtime changed to have this run only once during first need?

llvm/include/llvm/ADT/Triple.h
keryell: > also I am not sure there is any value to having fpga32 since to my understanding fpga don't have fixed bit width like traditional hardware.

I think this is related to the pointer size only. So we could have FPGA with less than GiB or more than 4GiB.
@aisoard: do we have HLS running with 32-bit addresses by the way?

sycl/test/lit.cfg.py
stsoe: @keryell XRT is multi-process safe, maybe I am not understand the comment correctly?

keryell: Thanks @stsoe.
So if XRT is multi-process-safe, I guess we should be able to run all the unit tests on our U200 in parallel without any failure since I assume there would be some kind of serialization around the access of the U200...
Curious. Either a bug in the SYCL runtime or in XRT. @Ralender can you dive a little bit in what happens here?

stsoe: @keryell .  XRT works closely like described here: https://xilinx.github.io/XRT/master/html/multiprocess.html.
I am not sure what you mean by serialization?  Maybe you will be surprised to know:

A process P2 will fail to download a new xclbin if some other process has the current xclbin locked.  The process P2 will fail, it will not be serialized by XRT.  

A process P1 will fail to acquire a lock on a CU after successful xclbin download, if in the mean time another process P2 has successfully downloaded a new xclbin.   P1 will fail, it will not wait for P2 to finish.

Of course there is serialization when all processes agree on the current xclbin and all processes agree to share at least some of the compute units such that no one gets left behind with nothing to execute on.

keryell: @stsoe Thanks! This is super helpful! That makes sense to have such an execution model on FPGA.
Obvious in our case all the unit tests runs different kernels, so the solution of @Ralender makes sense but the comment in the code above has to be updated by describing a little bit the XRT execution model here.
We probably need to think about extending the SYCL runtime with this knowledge at some point.
And obviously read again the XRT documentation from time to time to be aware of the new features... :-)

sycl/source/detail/program_manager/program_manager.cpp
Ralender: what the previous assert and now the throw detect is if there is multiple images inside the same executable that target different XCL_EMULATION_MODE, aka we do not support have some kernels in sw_emu and others in hw_emu.
and it would be quite hard since the emulation mode is global.

sycl/test/lit.cfg.py
Ralender: in https://xilinx.github.io/XRT/master/html/multiprocess.html
it is said "Emulation flow does not have support for multi-process yet."
and tests have to support emulation. 
it is very possible that even if XRT supports multi-process in hw mode.
XRT will report error when there is conflict on a resource and SYCL but there is no code to handle that error which causes test failure.

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,47 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
...
+  constexpr const char *env_var = "XCL_EMULATION_MODE";
+  if (strncmp("fpga64_", target, 7) == 0 ||
+      strncmp("fpga32_", target, 7) == 0) {
+    target += 7;
+    if (strcmp(target, "hw") == 0) {
+      if (has_been_invoked)
+        if (const char *mode = std::getenv(env_var))
+          throw sycl::platform_error(
+              std::string("all kernels are not compiled for the same "
+                          "environemnt. hw and ") +
```
keryell:
```suggestion
                          "environment: hw and ") +
```

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,47 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
...
+static void setupEnvironmentForKernels(RTDeviceBinaryImage *Img) {
+  static bool has_been_invoked = false;
+  const char* target = Img->getTarget();
+  constexpr const char *env_var = "XCL_EMULATION_MODE";
+  if (strncmp("fpga64_", target, 7) == 0 ||
+      strncmp("fpga32_", target, 7) == 0) {
+    target += 7;
+    if (strcmp(target, "hw") == 0) {
+      if (has_been_invoked)
+        if (const char *mode = std::getenv(env_var))
```
keryell, line 897: Is it possible for a user to set the variable to `hw` just to have a funny throw message below? :-)

sycl/source/detail/program_manager/program_manager.cpp
```diff
@@ -883,6 +883,47 @@ createKernelArgMask(const pi::ByteArray &Bytes) {
   return Result;
 }
 
+/// This will setup environment variable for XRT to match the type of kernels
+/// being loaded.
+static void setupEnvironmentForKernels(RTDeviceBinaryImage *Img) {
+  static bool has_been_invoked = false;
+  const char* target = Img->getTarget();
+  constexpr const char *env_var = "XCL_EMULATION_MODE";
```
keryell:
```suggestion
  constexpr auto env_var = "XCL_EMULATION_MODE";
```

sycl/test/lit.cfg.py
```diff
@@ -200,13 +203,24 @@ def getDeviceCount(device_type):
     lit_config.warning("Accelerator device not found")
 
 if xocc != "off":
-    xrt_lock="/tmp/xrt.lock"
-    acc_run_substitute+= "setsid flock -x " + xrt_lock + " "
+    # xrt doesn't deal well with multiple executable using it concurrently(at the time of writing).
```
keryell:
```suggestion
    # xrt doesn't deal well with multiple executables using it concurrently (at the time of writing).
```

sycl/test/lit.cfg.py
```diff
@@ -200,13 +203,24 @@ def getDeviceCount(device_type):
...
+    # the detail are at https://xilinx.github.io/XRT/master/html/multiprocess.html
```
keryell:
```suggestion
    # The details are at https://xilinx.github.io/XRT/master/html/multiprocess.html
```

sycl/test/lit.cfg.py
```diff
@@ -200,13 +203,24 @@ def getDeviceCount(device_type):
...
+    # so we wrap every use of XRT inside an file lock
+    # we also wrap invocation of executable in an setsid to prevent
```
keryell:
```suggestion
    # so we wrap every use of XRT inside an file lock.
    # We also wrap invocation of executable in an setsid to prevent
```

sycl/test/lit.cfg.py
```diff
@@ -200,13 +203,24 @@ def getDeviceCount(device_type):
...
+    acc_run_substitute+= "setsid flock --exclusive " + xrt_lock + " "
     if os.path.exists(xrt_lock):
         os.remove(xrt_lock)
-    if "XCL_EMULATION_MODE" in os.environ:
-        if os.environ["XCL_EMULATION_MODE"] == "hw":
-            acc_run_substitute="env -u XCL_EMULATION_MODE " + acc_run_substitute
+    # XCL_EMULATION_MODE = hw is only valid for our SYCL driver, 
+    # for XRT XCL_EMULATION_MODE should only be used when using either hw_emu or sw_emu
+    if xocc_target == "hw":
+        acc_run_substitute="env -u XCL_EMULATION_MODE " + acc_run_substitute
```
keryell:
```suggestion
        acc_run_substitute="env --unset=XCL_EMULATION_MODE " + acc_run_substitute
```
Explicit programming

sycl/source/detail/program_manager/program_manager.cpp
Ralender: no because all throws can happens only when has_been_invoked and during the first pass through the variable will be changed to match what the first kernel needs.

keryell:
Thanks. Quite an improvement in user experience!

PR #89, opened 2020-10-07
Fix edgedetection2

Ralender:
I thought edge detection was fixed but I saw issue with so here is a fix.
it dependes on https://github.com/triSYCL/sycl/pull/88 
because this patch allows every part of the toolchain be be aware of the target (sw_emu/hw_emu/hw)


keryell (commented):
I am waiting for the https://github.com/triSYCL/sycl/pull/88 to land first and the PR being rebased on top of it because it is unclear about what is the pure part of this PR.

sycl/include/CL/sycl/xilinx/fpga/partition_array.hpp
```diff
@@ -52,7 +52,10 @@ namespace partition {
   /// This fuction is currently empty but the LowerSYCLMetaData Pass will fill
   /// it with the required IR.
   template<typename Ptr>
-  __SYCL_DEVICE_ANNOTATE("xilinx_partition_array") __attribute__((always_inline))
+#if defined(__SYCL_XILINX_HW_EMU_MODE__) || defined(__SYCL_XILINX_HW_MODE__)
+  __SYCL_DEVICE_ANNOTATE("xilinx_partition_array")
+#endif
+  __attribute__((always_inline))
```
keryell: Would this compile with MSVC?
It would be nice to have this work in plain CPU mode.

clang/lib/Frontend/InitPreprocessor.cpp
```diff
@@ -1117,8 +1117,22 @@ static void InitializePredefinedMacros(const TargetInfo &TI,
...
       Builder.defineMacro("__SYCL_SPIR_DEVICE__");
+      switch (TI.getTriple().getSubArch()) {
+      case llvm::Triple::FPGASubArch_sw_emu:
+        Builder.defineMacro("__SYCL_XILINX_SW_EMU_MODE__");
+        break;
+      case llvm::Triple::FPGASubArch_hw_emu:
+        Builder.defineMacro("__SYCL_XILINX_HW_EMU_MODE__");
+        break;
+      case llvm::Triple::FPGASubArch_hw:
+        Builder.defineMacro("__SYCL_XILINX_HW_MODE__");
```
keryell, line 1130: These macros seems very useful.
They should be described somewhere in the documentation.

sycl/include/CL/sycl/xilinx/fpga/partition_array.hpp
Ralender: > Would this compile with MSVC?

I don't think so because the __ attribute format isn't supported by msvc as far AFAIK.
> It would be nice to have this work in plain CPU mode.

yes on gcc/clang because __SYCL_DEVICE_ANNOTATE will disappear.

keryell: Using a macro like `__SYCL_ALWAYS_INLINE` or `__TRISYCL_ALWAYS_INLINE`?

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -55,9 +59,39 @@ struct PrepareSYCLOpt : public ModulePass {
...
+      Instruction* GEP = GetElementPtrInst::Create(nullptr, Replacement, {Zero});
+      GEP->insertAfter(Cast);
+      Ptr.set(GEP);
+    }
+  }
+
   bool runOnModule(Module &M) override {
     turnNonKernelsIntoPrivate(M);
     setCallingConventions(M);
+    fixArrayPartition(M);
```
keryell: Is "fix" the right name?

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -55,9 +59,39 @@ struct PrepareSYCLOpt : public ModulePass {
     }
   }
 
+  void fixArrayPartition(Module& M) {
```
keryell: Add a description comment describing what is done from a high-level point of view.

keryell (changes requested):

sycl/include/CL/sycl/xilinx/fpga/partition_array.hpp
Ralender: ALWAYS_INLINE is the macro intel added so even if we create a new one we are not going to be shielded from collision on there code.

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -55,9 +59,45 @@ struct PrepareSYCLOpt : public ModulePass {
     }
   }
 
+  /// This will change array partition such that after the O3 pipeline it
+  /// matched very closely what v++ generates.
+  /// This will change the type of the alloca referenced by the array partition
+  /// into an array.
+  /// and change the argument received by xlx_array_partition into a pointer on
```
keryell: Fuse these lines or change the typesetting.

keryell: It would be interesting to explain which transformation is actually done here.

sycl/include/CL/sycl/xilinx/fpga/partition_array.hpp
```diff
@@ -52,7 +52,10 @@ namespace partition {
   /// This fuction is currently empty but the LowerSYCLMetaData Pass will fill
   /// it with the required IR.
   template<typename Ptr>
-  __SYCL_DEVICE_ANNOTATE("xilinx_partition_array") __attribute__((always_inline))
+#if defined(__SYCL_XILINX_HW_EMU_MODE__) || defined(__SYCL_XILINX_HW_MODE__)
+  __SYCL_DEVICE_ANNOTATE("xilinx_partition_array")
+#endif
+  ALWAYS_INLINE
```
keryell: The name has changed up-stream.

keryell (changes requested):

keryell (changes requested):
I have the feeling this has to be rebased on a more recent version to resolve the merge conflicts.

keryell (approved):
Thanks!

PR #90, opened 2020-10-07
[SYCL][NFC] Refactor unique name generation

Ralender:
this patch brings 2 good improvements:
 - stop using boost
 - factorize code for generation of unique name.

### DIFF
--- a/clang/lib/Sema/SemaSYCL.cpp
+++ b/clang/lib/Sema/SemaSYCL.cpp
@@ -27,6 +27,7 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/SHA1.h"
 
 #include <array>
 #include <functional>
@@ -740,6 +741,54 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
       AccTy->getTemplateArgs()[3].getAsIntegral().getExtValue());
 }
 
+/// Compute a unique name that is consumable by sycl-xocc
+static std::string computeUniqueSYCLXOCCName(StringRef Name,
+                                             StringRef Demangle) {
+  /// XOCC has a maximum of 64 character for the name of the kernel function
+  /// plus the name of one parameter.
+  /// Those characters need to be used wisely to prevent name collisions.
+  /// It is also useful to use a name that is understandable by the user,
+  /// so we add only 8 character of hash and only if needed.
+  constexpr unsigned MaxXOCCSize = 30;
+
+  std::string Result;
+  Result.reserve(Demangle.size());
+
+  for (char c : Demangle) {
+    if (!isAlphanumeric(c)) {
+      // Do not repeat _ in the cleaned-up name
+      if (!Result.empty() && Result.back() == '_')
+        continue;
+      c = '_';
+    }
+    Result.push_back(c);
+  }
+
+  /// The name alone is guaranteed to be unique, so if fits in the size, it is
+  /// enough.
+  if (Result.size() < MaxXOCCSize)
+    return Result;
+
+  /// 9 for 8 characters of hash and an '_'.
+  Result.erase(0, Result.size() - (MaxXOCCSize - 9));
+
+  /// Sadly there is only 63 valid characters in C identifiers.
+  /// So one of them A is repeated. This doesn't hurt entropy to much because
+  /// it is just 1 out of 64.
+  Result += '_' + llvm::SHA1::hashToString(
+                      llvm::ArrayRef<uint8_t>{
+                          reinterpret_cast<const uint8_t *>(Name.data()),
+                          Name.size()},
+                      "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+                      "abcdefghijklmnopqrstuvwxyz"
+                      "0123456789_A");
+
+  if (Result.size() > MaxXOCCSize)
+    Result.resize(MaxXOCCSize);
+
+  return Result;
+}
+
 // The first template argument to the kernel caller function is used to identify
 // the kernel itself.
 static QualType calculateKernelNameType(ASTContext &Ctx,
@@ -763,7 +812,11 @@ constructKernelName(Sema &S, FunctionDecl *KernelCallerFunc,
 
   MC.mangleTypeName(KernelNameType, Out);
 
-  return {std::string(Out.str()),
+  std::string Res = std::string(Out.str());
+  if (S.getASTContext().getTargetInfo().getTriple().isXilinxSYCLDevice())
+    Res = computeUniqueSYCLXOCCName(Res, KernelNameType.getAsString());
+
+  return {Res,
           PredefinedExpr::ComputeName(S.getASTContext(),
                                       PredefinedExpr::UniqueStableNameType,
                                       KernelNameType)};
--- a/llvm/lib/Support/SHA1.cpp
+++ b/llvm/lib/Support/SHA1.cpp
@@ -307,3 +307,27 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
   memcpy(Arr.data(), S.data(), S.size());
   return Arr;
 }
+
+std::string SHA1::hashToString(ArrayRef<uint8_t> Data, StringRef Mapping) {
+  constexpr auto BitsPerChar = 6;
+  constexpr auto CombinationPerChar = 1 << BitsPerChar;
+  (void)CombinationPerChar;
+  assert(Mapping.size() >= CombinationPerChar);
+  std::array<uint8_t, 20> Hash = SHA1::hash(Data);
+  std::string Res;
+  constexpr auto OutputLength =
+      (Hash.size() * 8 + BitsPerChar - 1) / BitsPerChar;
+  Res.reserve(OutputLength);
+
+  uint16_t Value = Hash[0];
+  for (unsigned AvailableBits = 8, Elem = 1; Res.size() < OutputLength;) {
+    Res.push_back(Mapping[Value & 0b111111]);
+    AvailableBits -= BitsPerChar;
+    Value >>= BitsPerChar;
+    if (AvailableBits < BitsPerChar && Elem < Hash.size()) {
+      Value |= Hash[Elem++] << AvailableBits;
+      AvailableBits += 8;
+    }
+  }
+  return Res;
+}
--- a/llvm/include/llvm/SYCL/InSPIRation.h
+++ b/llvm/include/llvm/SYCL/InSPIRation.h
@@ -17,8 +17,10 @@
 
 #include "llvm/IR/Module.h"
 #include "llvm/Pass.h"
+#include "llvm/Support/SHA1.h"
 
 namespace llvm {
+
 
 ModulePass *createInSPIRationPass();
 
--- a/llvm/include/llvm/Support/SHA1.h
+++ b/llvm/include/llvm/Support/SHA1.h
@@ -17,6 +17,7 @@
 
 #include <array>
 #include <cstdint>
+#include <string>
 
 namespace llvm {
 template <typename T> class ArrayRef;
@@ -50,6 +51,9 @@ public:
 
   /// Returns a raw 160-bit SHA1 hash for the given data.
   static std::array<uint8_t, 20> hash(ArrayRef<uint8_t> Data);
+
+  /// Returns a hash of the input represented using the provided maping.
+  static std::string hashToString(ArrayRef<uint8_t> Data, StringRef Mapping);
 
 private:
   /// Define some constants.
--- a/llvm/lib/SYCL/InSPIRation.cpp
+++ b/llvm/lib/SYCL/InSPIRation.cpp
@@ -38,30 +38,6 @@
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
 
-#define BOOST_NO_EXCEPTIONS
-
-/// \todo: Perhaps BOOST should appropriately be included through cmake. At the
-/// moment it's found via existing in the environment I think.. seems a little
-/// unclear at least. This goes for the run-time as well.
-#include <boost/container_hash/hash.hpp> // uuid_hasher
-#include <boost/uuid/uuid_generators.hpp> // sha name_gen/generator
-#include <boost/uuid/uuid_io.hpp> // uuid to_string
-#include <boost/version.hpp>
-
-// BOOST_NO_EXCEPTIONS enabled so we need to define our own throw_exception or
-// get a linker error.
-namespace boost {
-#if BOOST_VERSION < 107300
-void throw_exception(std::exception const &e) {
-  llvm_unreachable("exception are disabled");
-}
-#else
-void throw_exception(std::exception const &e, boost::source_location const&) {
-  llvm_unreachable("exception are disabled");
-}
-#endif
-} // namespace boost
-
 using namespace llvm;
 
 // Put the code in an anonymous namespace to avoid polluting the global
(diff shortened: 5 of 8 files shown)


clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,53 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
 
+/// Compute a unique name that is consumable by sycl-xocc
+static std::string computeUniqueSYCLXOCCName(StringRef Name,
+                                             StringRef Demangle) {
+  /// XOCC has a maximum of 64 character for the name of the kernel function
+  /// plus the name of one parameter.
+  /// Those caracters need to be used wisely to prevent name collisions.
+  /// Its is also useful to use a name that is understandable nby the user.
+  /// so we add only 8 character of hash an only if needed.
+  constexpr unsigned MaxXOCCSize = 30;
```
keryell, line 752:
```suggestion
  constexpr std::size_t MaxXOCCSize = 30;
```
do not use old lazy C style.

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,53 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
+
+  std::string Result;
+
+  for (char c : Demangle) {
+    if (c == ':' || c == ' ' || c == '<' || c == '>' || c == ',') {
```
keryell: Would it be possible to have also `[]()*&` in the naming path?
```c++
void weird(int* (&arg)[10]) {
  // launch a lambda kernel here...
}
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,53 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
 }
 
+/// Compute a unique name that is consumable by sycl-xocc
+static std::string computeUniqueSYCLXOCCName(StringRef Name,
+                                             StringRef Demangle) {
+  /// XOCC has a maximum of 64 character for the name of the kernel function
+  /// plus the name of one parameter.
+  /// Those caracters need to be used wisely to prevent name collisions.
+  /// Its is also useful to use a name that is understandable nby the user.
+  /// so we add only 8 character of hash an only if needed.
```
keryell:
```suggestion
  /// XOCC has a maximum of 64 character for the name of the kernel function
  /// plus the name of one parameter.
  /// Those characters need to be used wisely to prevent name collisions.
  /// It is also useful to use a name that is understandable by the user,
  /// so we add only 8 character of hash and only if needed.
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,53 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
+
+  std::string Result;
+
+  for (char c : Demangle) {
+    if (c == ':' || c == ' ' || c == '<' || c == '>' || c == ',') {
+      if (!Result.empty() && Result.back() == '_')
+        continue;
+      c = '_';
+    }
+    if (c == '$')
```
keryell: `else if` just for guiding the reader

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,53 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
+  /// Those caracters need to be used wisely to prevent name collisions.
+  /// Its is also useful to use a name that is understandable nby the user.
+  /// so we add only 8 character of hash an only if needed.
+  constexpr unsigned MaxXOCCSize = 30;
+
+  std::string Result;
+
+  for (char c : Demangle) {
+    if (c == ':' || c == ' ' || c == '<' || c == '>' || c == ',') {
+      if (!Result.empty() && Result.back() == '_')
```
keryell, line 760:
```suggestion
      // Do not repeat _ in the cleaned-up name
      if (!Result.empty() && Result.back() == '_')
```
thanks for the quizz!

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,53 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
+        continue;
+      c = '_';
+    }
+    if (c == '$')
+      c = 'S';
+    Result.push_back(c);
+  }
+
+  /// Name alone is garanteed to be unique so if fits in the size. it is enough.
```
keryell:
```suggestion
  /// Name alone is guaranteed to be unique, so if fits in the size, it is enough.
```

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,39 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
   memcpy(Arr.data(), S.data(), S.size());
   return Arr;
 }
+
+std::string SHA1::hashToString(ArrayRef<uint8_t> Data, StringRef Mapping) {
+  constexpr unsigned BitsPerChar = 6;
```
keryell: Replace all the `unsigned` by `unsigned int` or `std::size_t`

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,39 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+                     << LShift1[Pos];
+    uint8_t Value2 = 0;
+    if (LLVM_LIKELY(Elem + 1 < Hash.size()))
+      Value2 = (Hash[Elem + 1] & Mask2[Pos]) >> RShift2[Pos];
+    Res.push_back(Mapping[Value1 | Value2]);
+    Bits += BitsPerChar;
+    if (Bits >= 8) {
+      Elem++;
+      Bits -= 8;
+    }
```
keryell, line 330: Good obfuscation exercise!
Explain the FSM.
Not clear that at the end all the bits are consumed...
It would be simpler with a `std::range` of bits... :-)
Otherwise I wonder it would not be more understandable with a https://en.cppreference.com/w/cpp/utility/bitset `std::bitset<16>` that you fill by the left with `... | std::bitset<16> { c } << 8` when 1 char has been consumed and you empty it to the right with a `(... & std::bitset<16> { (1 << 6) - 1 }).to_ulong()`, then `>>= 6` and loop.

keryell (changes requested):

llvm/lib/Support/SHA1.cpp
AlexeySachkov: Hi,

Just curious, what is wrong with `unsigned` compared to `unsigned int`? Why the latter is preferred over the former?

llvm/lib/Support/SHA1.cpp
Ralender: > Good obfuscation exercise!
this wasn't the intent.

> Explain the FSM.
> Not clear that at the end all the bits are consumed...

I added assert at the end of the loop for that purpose.
> It would be simpler with a `std::range` of bits... :-)

this part of the code is build in c++14 mode so no std::range
and i would need index any way. to access the next element.
> Otherwise I wonder it would not be more understandable with a https://en.cppreference.com/w/cpp/utility/bitset `std::bitset<16>` that you fill by the left with `... | std::bitset<16> { c } << 8` when 1 char has been consumed and you empty it to the right with a `(... & std::bitset<16> { (1 << 6) - 1 }).to_ulong()`, then `>>= 6` and loop.

keryell: These asserts do not explain anything.
After sleeping over this, I think that just having a `std::uint16_t` to play the role of accumulator/extractor and just using shift and mask to fill 8 bits and extract 6 bits are enough. No need for `std::bitset`.
This should be quite simpler and more understandable than the FSM approach.
Yes, I know that Xilinx is a hardware company, but writing FPGA RTL-like C++ code might be a little bit excessive. :-)

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,46 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
   memcpy(Arr.data(), S.data(), S.size());
   return Arr;
 }
+
+std::string SHA1::hashToString(ArrayRef<uint8_t> Data, StringRef Mapping) {
+  constexpr unsigned BitsPerChar = 6;
+  constexpr unsigned CombinationPerChar = 1 << BitsPerChar;
+  constexpr unsigned LoopIterations = 27;
```
keryell: Why? Put an expression.
I have the feeling that just using `auto` everywhere is enough instead of `unsigned` all over the place.

llvm/lib/Support/SHA1.cpp
keryell: Just to be more explicit, for example for the Python programmers that could drop by. :-)

keryell (changes requested):

llvm/lib/Support/SHA1.cpp
Ralender: using uint16_t greatly improved readability

llvm/lib/Support/SHA1.cpp
Ralender: i am not a fan of using auto in this context.

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -745,6 +746,54 @@ static target getAccessTarget(const ClassTemplateSpecializationDecl *AccTy) {
...
+  /// Those characters need to be used wisely to prevent name collisions.
+  /// It is also useful to use a name that is understandable by the user,
+  /// so we add only 8 character of hash and only if needed.
+  constexpr unsigned MaxXOCCSize = 30;
+
+  std::string Result;
+  Result.reserve(Demangle.size());
+
+  for (char c : Demangle) {
+    if (!isAlphanumeric(c)) {
```
keryell, line 758: Quite simpler!

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -768,7 +817,11 @@ constructKernelName(Sema &S, FunctionDecl *KernelCallerFunc,
 
   MC.mangleTypeName(KernelNameType, Out);
 
-  return {std::string(Out.str()),
+  std::string Res = std::string(Out.str());
```
keryell, line 815:
```suggestion
  auto Res = Out.str();
```

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+    Res.push_back(Mapping[(Value & Mask) >> Shift]);
+    Bits += BitsPerChar;
+    if (Bits >= 8) {
+      Elem++;
+      Bits -= 8;
+    }
+    if (Elem >= Hash.size())
+      break;
+  }
+  assert(Elem == 20 && Bits == 2 && Res.size() == LoopIterations);
```
keryell: Why 20 and 2?
`Elem == std::size(Hash)` but still figuring out for 2...

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+    uint16_t Shift = 10 - Bits;
+    uint16_t Mask = 0b111111 << Shift;
+    Value = (Value & Mask) >> Shift;
+    Res.push_back(Mapping[(Value & Mask) >> Shift]);
+    Bits += BitsPerChar;
+    if (Bits >= 8) {
+      Elem++;
+      Bits -= 8;
+    }
+    if (Elem >= Hash.size())
```
keryell: Replace the while with `for(unsigned Bits = 0, Elem = 0;; Elem < Hash.size())`

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+  constexpr auto LoopIterations = 27;
+  assert(Mapping.size() >= CombinationPerChar);
+  std::array<uint8_t, 20> Hash = SHA1::hash(Data);
+  std::string Res;
+  Res.reserve(LoopIterations);
+
+  unsigned Bits = 0;
+  unsigned Elem = 0;
+  while (1) {
+    uint16_t Value = Hash[Elem] << 8;
```
keryell: Too complex...
This should be in the loop prologue, before the loop.
No need to reload every time this.
You can remove a lot of shifts if you load the new bits by the MSB and extract by LSB.
```c++
uint16_t Value = Hash[Elem];
```

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+    if (LLVM_LIKELY(Elem + 1 < Hash.size()))
+      Value |= Hash[Elem + 1];
```
keryell:
```suggestion
      Value |= Hash[Elem + 1] << 8;
```
also you need to update the `if` condition to do it only when you have consumed all the 8 MSB bit.

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+
+    uint16_t Shift = 10 - Bits;
```
keryell: All this is still quite too complex :-)

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+    uint16_t Mask = 0b111111 << Shift;
+    Value = (Value & Mask) >> Shift;
+    Res.push_back(Mapping[(Value & Mask) >> Shift]);
```
keryell: At the end, if we use the LSB, this is just
```suggestion
    Res.push_back(Mapping[Value & ((1 << BitsPerChar) - 1)]);
```
and the 2 previous lines are removed

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,36 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+    Bits += BitsPerChar;
+    if (Bits >= 8) {
+      Elem++;
+      Bits -= 8;
```
keryell: OK, so this is actually where you should have
```c++
      Value |= Hash[Elem] << 8;
```

keryell (changes requested):
Quite better but still far from the zen understandable code :-)

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,27 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
+  for (unsigned Bits = 0, Elem = 1; Res.size() < LoopIterations;) {
+    Res.push_back(Mapping[Value & 0b111111]);
+    Bits += BitsPerChar;
+    Value >>= BitsPerChar;
+    if (Bits >= 8) {
+      Elem++;
+      Bits -= 8;
+      if (LLVM_LIKELY(Elem < Hash.size()))
+        Value |= Hash[Elem] << (8 - Bits);
+    }
```
keryell: I found this still difficult to understand because `Bits` which is not physical. What about:
```suggestion
  uint16_t Value = Hash[0];
  for (auto AvailableBits = CHAR_BITS, Elem = 1; AvailableBits > 0;) {
    Res.push_back(Mapping[Value & ((1 << BitsPerChar) - 1)]);
    AvailableBits -= BitsPerChar;
    Value >>= BitsPerChar;
    if (AvailableBits < BitsPerChar && Elem < Hash.size()) {
        Value |= Hash[Elem++] << AvailableBits;
        AvailableBits += CHAR_BITS;
    }
```
I think now we have reached the bottom line here.
This is a rehearsal for your future coding interviews... :-)

llvm/lib/Support/SHA1.cpp
```diff
@@ -307,3 +307,27 @@ std::array<uint8_t, 20> SHA1::hash(ArrayRef<uint8_t> Data) {
...
 }
+
+std::string SHA1::hashToString(ArrayRef<uint8_t> Data, StringRef Mapping) {
+  constexpr auto BitsPerChar = 6;
+  constexpr auto CombinationPerChar = 1 << BitsPerChar;
+  constexpr auto LoopIterations = 27;
+  assert(Mapping.size() >= CombinationPerChar);
+  std::array<uint8_t, 20> Hash = SHA1::hash(Data);
+  std::string Res;
+  Res.reserve(LoopIterations);
```
keryell:
```suggestion
  assert(Mapping.size() >= CombinationPerChar);
  std::array<uint8_t, 20> Hash = SHA1::hash(Data);
  std::string Res;
  constexpr auto OutputLength = (Hash.size()*CHARBITS + BitsPerChar - 1)/BitsPerChar;
  Res.reserve(OutputLength);
```

keryell (changes requested):
Still some clarifications & simplifications

keryell (approved):
Great simplification!

PR #92, opened 2020-10-14
Add DDR banks

Ralender:
This depends on #91 because Intel's implementation was not complete in the previous version we were based on.

### DIFF
--- a/clang/lib/Sema/SemaSYCL.cpp
+++ b/clang/lib/Sema/SemaSYCL.cpp
@@ -1613,6 +1613,25 @@ public:
   }
 };
 
+static bool isSyclXilinxType(const QualType &Ty) {
+  static std::array<StringRef, 3> Namespaces = {"cl", "sycl", "xilinx"};
+  llvm::SmallVector<const DeclContext *, 8> CtxStack;
+  CtxStack.push_back(cast<DeclContext>(Ty->getAsTagDecl()));
+  while (!isa<TranslationUnitDecl>(CtxStack.back()->getParent()))
+    CtxStack.push_back(CtxStack.back()->getParent());
+  for (unsigned Idx = 0; Idx < Namespaces.size(); Idx++) {
+    auto *NS = dyn_cast<NamespaceDecl>(CtxStack.pop_back_val());
+    if (!NS)
+      return false;
+    IdentifierInfo *II = NS->getIdentifier();
+    if (!II)
+      return false;
+    if (!II->isStr(Namespaces[Idx]))
+      return false;
+  }
+  return true;
+}
+
 // A type to Create and own the FunctionDecl for the kernel.
 class SyclKernelDeclCreator : public SyclKernelFieldHandler {
   FunctionDecl *KernelDecl;
@@ -1649,6 +1668,33 @@ class SyclKernelDeclCreator : public SyclKernelFieldHandler {
     Params.push_back(NewParam);
   }
 
+  // Obtain an integer value stored in a template parameter of buffer_location
+  // property to pass it to buffer_location kernel attribute
+  void handleXilinxProperty(ParmVarDecl *Param, QualType PropTy,
+                            SourceLocation Loc) {
+    if (!isSyclXilinxType(PropTy))
+      return;
+    /// TODO: when D88645 lands update this code to use that instead.
+    ASTContext &Ctx = SemaRef.getASTContext();
+    const CXXRecordDecl *RD = PropTy->getAsCXXRecordDecl();
+    const CXXRecordDecl *PRD = cast<CXXRecordDecl>(RD->getParent());
+    std::string Args;
+    if (const auto *PropDecl = dyn_cast<ClassTemplateSpecializationDecl>(RD)) {
+      for (auto &Arg : PropDecl->getTemplateArgs().asArray()) {
+        switch (Arg.getKind()) {
+        case TemplateArgument::Integral:
+          Args += "_" + std::to_string(static_cast<int>(
+                            Arg.getAsIntegral().getExtValue()));
+          break;
+        default:
+          llvm_unreachable("unimplemented");
+        }
+      }
+    }
+    Param->addAttr(AnnotateAttr::CreateImplicit(
+        Ctx, "xilinx_" + PRD->getName().str() + Args));
+  }
+
   // Handle accessor properties. If any properties were found in
   // the accessor_property_list - add the appropriate attributes to ParmVarDecl.
   void handleAccessorPropertyList(ParmVarDecl *Param,
@@ -1669,6 +1715,7 @@ class SyclKernelDeclCreator : public SyclKernelFieldHandler {
       QualType PropTy = Prop->getAsType();
       if (Util::isSyclBufferLocationType(PropTy))
         handleBufferLocationProperty(Param, PropTy, Loc);
+      handleXilinxProperty(Param, PropTy, Loc);
     }
   }
 
--- a/llvm/lib/SYCL/PrepareSYCLOpt.cpp
+++ b/llvm/lib/SYCL/PrepareSYCLOpt.cpp
@@ -17,6 +17,7 @@
 
 #include "llvm/IR/CallingConv.h"
 #include "llvm/IR/Instructions.h"
+#include "llvm/IR/Intrinsics.h"
 #include "llvm/SYCL/PrepareSYCLOpt.h"
 #include "llvm/Support/Casting.h"
 
@@ -55,9 +56,26 @@ struct PrepareSYCLOpt : public ModulePass {
     }
   }
 
+  /// At this point in the pipeline Annotations intrinsic have all been
+  /// converted into what they need to be. But they can still be present and
+  /// have pointer on pointer as arguments which v++ can't deal with.
+  void removeAnnotationsIntrisic(Module &M) {
+    SmallVector<Instruction *, 16> ToRemove;
+    for (Function &F : M.functions())
+      if (F.getIntrinsicID() == Intrinsic::annotation ||
+          F.getIntrinsicID() == Intrinsic::ptr_annotation ||
+          F.getIntrinsicID() == Intrinsic::var_annotation)
+        for (User *U : F.users())
+          if (auto *I = dyn_cast<Instruction>(U))
+            ToRemove.push_back(I);
+    for (Instruction *I : ToRemove)
+      I->eraseFromParent();
+  }
+
   bool runOnModule(Module &M) override {
     turnNonKernelsIntoPrivate(M);
     setCallingConventions(M);
+    removeAnnotationsIntrisic(M);
     return true;
   }
 };
--- a/sycl/include/CL/sycl/xilinx/fpga/memory_properties.hpp
+++ b/sycl/include/CL/sycl/xilinx/fpga/memory_properties.hpp
@@ -1,0 +1,55 @@
+//==- memory_properties.hpp --- SYCL Xilinx memory proprerties       -------==//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef SYCL_XILINX_FPGA_MEMORY_PROPERTIES_HPP
+#define SYCL_XILINX_FPGA_MEMORY_PROPERTIES_HPP
+
+#include "CL/sycl/detail/defines.hpp"
+#include "CL/sycl/detail/property_helper.hpp"
+#include "CL/sycl/properties/accessor_properties.hpp"
+
+__SYCL_INLINE_NAMESPACE(cl) {
+
+namespace sycl {
+namespace xilinx {
+namespace property {
+
+struct ddr_bank {
+  template <unsigned A> struct instance {
+    template <int B> constexpr bool operator==(const instance<B> &) const {
+      return A == B;
+    }
+    template <int B> constexpr bool operator!=(const instance<B> &) const {
+      return A != B;
+    }
+  };
+};
+} // namespace property
+
+template <typename... Ts>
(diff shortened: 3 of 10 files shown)


keryell (commented):
As discussed today, I am not a fan of changing `SemaSYCL.cpp` just for this.
This sounds business as usual like the old ancient time of HLS C where there was no abstraction and so any problem had to be solved by keywords, `#pragma` and attributes...
In C++ this decoration should be handled by the SYCL header and metaprogramming, even if Intel is going old school by dumping their kitchen sink in the front-end. @mkinsner :-)

Ralender:
> As discussed today, I am not a fan of changing `SemaSYCL.cpp` just for this.
> This sounds business as usual like the old ancient time of HLS C where there was no abstraction and so any problem had to be solved by keywords, `#pragma` and attributes...

to clarify the API exposed to the user is pure C++, you can look at the added test to see how it looks like.
> In C++ this decoration should be handled by the SYCL header and metaprogramming, even if Intel is going old school by dumping their kitchen sink in the front-end. @mkinsner :-)

this doesn't fit well with the current implementation where the wrapper code generated around lambdas is generated by the front-ed and isn't part of the implementation.

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
```diff
@@ -52,12 +52,11 @@ void dataflow(T functor) {
     \param[in] f is a function with an innermost loop to be executed in a
     pipeline way.
 */
-template <typename T>
-__SYCL_DEVICE_ANNOTATE("xilinx_pipeline") __attribute__((always_inline))
-void pipeline(T functor) {
+template<typename T>
+__SYCL_DEVICE_ANNOTATE("xilinx_pipeline")
+ALWAYS_INLINE void pipeline(T functor) {
```
j-stephan: Why not use `[[gnu::always_inline]]` here? This would follow the standard C++ syntax and remove the need for using the preprocessor.

sycl/include/CL/sycl/xilinx/fpga/partition_array.hpp
```diff
@@ -52,8 +52,8 @@ namespace partition {
   /// This fuction is currently empty but the LowerSYCLMetaData Pass will fill
   /// it with the required IR.
   template<typename Ptr>
-  __SYCL_DEVICE_ANNOTATE("xilinx_partition_array") __attribute__((always_inline))
-  inline void xilinx_partition_array(Ptr, int, int, int) {}
+  __SYCL_DEVICE_ANNOTATE("xilinx_partition_array")
+  ALWAYS_INLINE inline void xilinx_partition_array(Ptr, int, int, int) {}
```
j-stephan, line 56: Same here.

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
keryell: I guess this is for handling any compiler, such as Microsoft Visual Studio?

j-stephan: That is the beauty of this attribute syntax: compilers will ignore any attributes they don't understand. The worst that can happen is a compiler warning about using an unknown attribute.

keryell:
> this doesn't fit well with the current implementation where the wrapper code generated around lambdas is generated by the front-ed and isn't part of the implementation.

Would it be possible to have a compromise with a more generic wrapper code independent from the exact properties? For example just a `handleXilinxProperties` that will trust blindly any kind of random property with the right prefix?
We can also postpone this when the next property shows up...
Otherwise we could also metaprogram the `constexpr` property handler to gather all the Xilinx properties under an Über-property type that would be tested in `SemaSYCL`? Probably more complex...

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
keryell: That is a good point. :-) Yes, the compiler might barf...

Ralender: one advantage of macro is that we can add support for multiple compiler, but i don't think we will.
I have still not tested the project on windows.

Ralender (commented):
> > this doesn't fit well with the current implementation where the wrapper code generated around lambdas is generated by the front-ed and isn't part of the implementation.
> 
> Would it be possible to have a compromise with a more generic wrapper code independent from the exact properties? 

you mean In the frontend ? the changes are already quite small, just add an annotation on accessor parameter of a kernel with a DDR bank specified.
> For example just a `handleXilinxProperties` that will trust blindly any kind of random property with the right prefix?

in the front-end this seems possible, i think the prefix would be the cl::sycl::xilinx namespace.
> We can also postpone this when the next property shows up...
> Otherwise we could also metaprogram the `constexpr` property handler to gather all the Xilinx properties under an Über-property type that would be tested in `SemaSYCL`? Probably more complex...

it might be possible but it is much harder to do than what we have here with at best no difference for the user.

Ralender:
added missing files.

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
keryell: Probably `ALWAYS_INLINE` is an unfortunate portable way to go.
I have the feeling it should have a more protected name, such as `__TRISYCL_ALWAYS_INLINE` or whatever.

keryell:
> you mean In the frontend ? the changes are already quite small, just add an annotation on accessor parameter of a kernel with a DDR bank specified.
> in the front-end this seems possible, i think the prefix would be the cl::sycl::xilinx namespace.
> it might be possible but it is much harder to do than what we have here with at best no difference for the user.

Yes, factorize as much as possible in the frontend work.
I would prefer an intensive solution to an extensive solution (enumerate each case both in the SYCL headers AND in the front-end).
Technical debt... :-)

Ralender:
> > you mean In the frontend ? the changes are already quite small, just add an annotation on accessor parameter of a kernel with a DDR bank specified.
> > in the front-end this seems possible, i think the prefix would be the cl::sycl::xilinx namespace.
> > it might be possible but it is much harder to do than what we have here with at best no difference for the user.
> 
> Yes, factorize as much as possible in the frontend work.
> I would prefer an intensive solution to an extensive solution (enumerate each case both in the SYCL headers AND in the front-end).
> Technical debt... :-)

I tried to minimize factorize front-end code as much as possible.
does this seems enough ?

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -55,9 +56,26 @@ struct PrepareSYCLOpt : public ModulePass {
     }
   }
 
+  /// At this point in the pipeline Annotations intrinsic have all been
+  /// converted into what they need to be. But they can still be present and
+  /// have pointer on pointer as arguments which XOCC can't deal with.
```
keryell:
```suggestion
  /// have pointer on pointer as arguments which v++ can't deal with.
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -1613,6 +1613,20 @@ class SyclKernelDecompMarker : public SyclKernelFieldHandler {
   }
 };
 
+static bool isSyclXilinxDDRBankType(const QualType &Ty) {
```
keryell: Is it possible to be more generic, such as with ` static bool isSyclXilinxType(const QualType &Ty) {` wher it can introspect the types and guess what to do from some carefully crafted types in the `.hpp`?

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -1649,6 +1663,20 @@ class SyclKernelDeclCreator : public SyclKernelFieldHandler {
...
+  void handleXilinxProperty(ParmVarDecl *Param, QualType PropTy,
+                             SourceLocation Loc) {
+    ASTContext &Ctx = SemaRef.getASTContext();
+    const auto *PropDecl =
+        cast<ClassTemplateSpecializationDecl>(PropTy->getAsRecordDecl());
+    const auto BufferLoc = PropDecl->getTemplateArgs()[0];
+    int LocationID = static_cast<int>(BufferLoc.getAsIntegral().getExtValue());
+    if (isSyclXilinxDDRBankType(PropTy))
+      Param->addAttr(AnnotateAttr::CreateImplicit(
+          Ctx, "xilinx_ddr_bank_" + std::to_string(LocationID)));
```
keryell: This could be generic.

sycl/include/CL/sycl/xilinx/fpga/memory_properties.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "CL/sycl/detail/property_helper.hpp"
+#include "CL/sycl/properties/accessor_properties.hpp"
+
+__SYCL_INLINE_NAMESPACE(cl) {
+
+namespace sycl {
+namespace xilinx {
+namespace property {
+
+struct ddr_bank {
```
keryell, line 22: This could be a higher-order self-describing property to be introspected by Clang to do what is required to be done...

keryell (changes requested):
This could be generalized to go towards you generalized annotations.
At the end, basically nothing in Clang. :-)

Ralender:
> This could be generalized to go towards you generalized annotations.
> At the end, basically nothing in Clang. :-)

the implementation was generalized even more by treating any type inside cl::sycl::xilinx as a valid property.
and will generate an annotate attribute of the form "xilinx_name of the class_args".
so for a property cl::sycl::xilinx::ddr_bank<1> it will generate "xilinx_ddr_bank_1"

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -1649,6 +1668,33 @@ class SyclKernelDeclCreator : public SyclKernelFieldHandler {
...
+  // Obtain an integer value stored in a template parameter of buffer_location
+  // property to pass it to buffer_location kernel attribute
+  void handleXilinxProperty(ParmVarDecl *Param, QualType PropTy,
+                            SourceLocation Loc) {
+    if (!isSyclXilinxType(PropTy))
+      return;
+    /// TODO: when D88645 lands update this code to use that instead.
+    ASTContext &Ctx = SemaRef.getASTContext();
+    const CXXRecordDecl *RD = PropTy->getAsCXXRecordDecl();
+    const CXXRecordDecl *PRD = cast<CXXRecordDecl>(RD->getParent());
```
keryell, line 1680: DRY:
```suggestion
    auto &Ctx = SemaRef.getASTContext();
    const auto RD = PropTy->getAsCXXRecordDecl();
    const auto PRD = cast<CXXRecordDecl>(RD->getParent());
```
anyway, does not need to make the change if it is throwaway code...

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
```diff
@@ -38,9 +39,9 @@ namespace sycl::xilinx {
 */
 
 template <typename T>
-void dataflow(T functor) {
+void dataflow(T&& functor) {
   _ssdm_op_SpecDataflowPipeline(-1, "");
-  functor();
+  std::forward<T>(functor)();
```
keryell, line 46: I agree about the `T&&` above but why do we need a `std::forward<T>` here since it is just to use it here and not to forward to another function?

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
```diff
@@ -17,6 +17,7 @@
 
 #include "CL/sycl/xilinx/fpga/ssdm_inst.hpp"
 #include "CL/sycl/detail/defines.hpp"
+#include <utility>
```
keryell, line 20: Related to other comment below.

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
```diff
@@ -52,12 +53,11 @@ void dataflow(T functor) {
...
     pipeline way.
 */
-template <typename T>
-__SYCL_DEVICE_ANNOTATE("xilinx_pipeline") __attribute__((always_inline))
-void pipeline(T functor) {
-  functor();
+template<typename T>
+__SYCL_DEVICE_ANNOTATE("xilinx_pipeline")
+ALWAYS_INLINE void pipeline(T&& functor) {
+  std::forward<T>(functor)();
```
keryell, line 63: Idem.
Otherwise I have submitted https://github.com/intel/llvm/issues/2665 about `ALWAYS_INLINE`.

sycl/test/xocc_tests/simple_tests/ddr_bank_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <CL/sycl.hpp>
+
+int main() {
+  cl::sycl::buffer<cl::sycl::cl_int, 1> Buffer(4);
+  cl::sycl::queue Queue;
+  cl::sycl::range<1> NumOfWorkItems{Buffer.get_count()};
+
+  Queue.submit([&](cl::sycl::handler &cgh) {
+    sycl::ONEAPI::accessor_property_list PL{sycl::xilinx::ddr_bank<1>};
```
keryell, line 14: Remove this intermediate variables.
Rationale: they are going to change their type soon and there will be a merge conflict.

keryell (changes requested):

sycl/include/CL/sycl/xilinx/fpga/opt_decorate_func.hpp
Ralender: the operatror() of the functor is an other function.

this can can change break user code if said user code is sufficiently weird.

example https://godbolt.org/z/cvzoPx

keryell: Ouch! This is right... I should be careful about the hidden forwarding...
Can you add a comment about why in both places?

keryell (approved):
Great!

PR #93, opened 2020-10-20
[SYCL] Fix kernel naming for hw_emu

Ralender:
hw_emu has higher requirements than sw_emu

### DIFF
--- a/clang/lib/Sema/SemaSYCL.cpp
+++ b/clang/lib/Sema/SemaSYCL.cpp
@@ -749,7 +749,13 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
   /// Those characters need to be used wisely to prevent name collisions.
   /// It is also useful to use a name that is understandable by the user,
   /// so we add only 8 character of hash and only if needed.
+  /// The first character cannot be an underscore or a digit.
+  /// An underscore can't be followed by an other underscore.
   constexpr unsigned MaxXOCCSize = 30;
+  /// Some transformations might make 2 kernel identifiers the same.
+  /// Allow adding a hash when such transformations are made to avoid possible
+  /// name conflict.
+  bool ForceHash = false;
 
   std::string Result;
   Result.reserve(Demangle.size());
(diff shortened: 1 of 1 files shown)


clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -749,7 +749,10 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
   /// Those characters need to be used wisely to prevent name collisions.
   /// It is also useful to use a name that is understandable by the user,
   /// so we add only 8 character of hash and only if needed.
+  /// The first character cannot be an underscore or a didgit.
```
keryell:
```suggestion
  /// The first character cannot be an underscore or a digit.
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -749,7 +749,10 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
...
+  /// An underscore can't be followed by an other underscore.
   constexpr unsigned MaxXOCCSize = 30;
+  bool ForceHash = false;
```
keryell, line 758:
```suggestion
  /// Some transformations might make 2 kernel identifiers the same.
  /// Allow adding a hash when such transformations are made to avoid possible name conflict.
  bool ForceHash = false;
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -764,18 +767,29 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
     Result.push_back(c);
   }
 
+  if (!Result.empty() && (Result.front() == '_' || isDigit(Result.front()))) {
```
keryell:
```suggestion
  // Replace first kernel character name by a 'k' to be compatible with SPIR
  if (!Result.empty() && (Result.front() == '_' || isDigit(Result.front()))) {
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -764,18 +767,29 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
...
+  }
+
   /// The name alone is guaranteed to be unique, so if fits in the size, it is
   /// enough.
-  if (Result.size() < MaxXOCCSize)
+  if (Result.size() < MaxXOCCSize && !ForceHash)
     return Result;
 
   /// 9 for 8 characters of hash and an '_'.
   Result.erase(0, Result.size() - (MaxXOCCSize - 9));
```
keryell, line 785: An easy way to simplify the following would be to suppress '_' from the hash string (62 characters among the 64) and start putting the hash at position 1 instead of 0. 
Because it is not impossible that the hash will be `________________` too...

keryell (changes requested):

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -764,24 +770,36 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
...
-                      "abcdefghijklmnopqrstuvwxyz"
-                      "0123456789_A");
+  if (!Result.empty() && (Result.front() == '_' || isDigit(Result.front())))
+    Result.front() = 'k';
+
+  if (!Result.empty() && Result.back() != '_')
+    Result.push_back('_');
+
+  /// Sadly there is only 63 valid characters in C identifiers and v++ doesn't
+  /// deal well with  double underscores in identifiers. So A and B are
```
keryell:
```suggestion
  /// deal well with double underscores in identifiers. So A and B are
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -764,24 +770,36 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
...
+  /// repeated. This doesn't hurt entropy to much because it is just 2 out
```
keryell:
```suggestion
  /// repeated. This doesn't hurt entropy too much because it is just 2 out
```

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -764,24 +770,36 @@ static std::string computeUniqueSYCLXOCCName(StringRef Name,
     Result.push_back(c);
   }
 
+  // Replace first kernel character name by a 'k' to be compatible with SPIR
+  if (!Result.empty() && (Result.front() == '_' || isDigit(Result.front()))) {
```
keryell: By looking at all these `Result.empty()` everywhere I wonder in which case this could be empty.

keryell (changes requested):
Getting close!

PR #96, opened 2020-11-06
[SYCL][CUDA][FPGA] allow running kernels on CUDA and FPGA in the same …

Ralender:
…application

-fsycl-xocc is removed since it doesn't fit well with mixing kernel targets


keryell (commented):
Thanks! This looks good.

PR #105, opened 2021-01-18
Pull down

Ralender:
Pull down intel's changes and fixes what they break.


keryell:
Usually the new versions of upstream come into `sycl/unified/next`.
Do you still want this is `sycl/unified/master`?

Ralender:
yeah this should be merged into sycl/unified/next i changed it.

keryell:
Thanks!

Issue #107, opened 2021-02-05
clang keywords break toolchain compilation with gcc


j-stephan:
I thought I'd give this compiler another try since #102 has been merged. With gcc the compilation of the toolchain itself breaks because there are clang-specific keywords in the source code. The offending line is here:

https://github.com/triSYCL/sycl/blob/b6f942716a40846fa02b548ee4217945f3e157cb/sycl/include/CL/sycl/detail/cstr.hpp#L55

gcc's error message:

```
In file included from /home/jan/sycl-workspace/sycl/sycl/include/CL/sycl/xilinx/fpga/kernel_properties.hpp:21,
                 from /home/jan/sycl-workspace/sycl/sycl/include/CL/sycl/handler.hpp:31,
                 from /home/jan/sycl-workspace/sycl/sycl/include/CL/sycl/ONEAPI/reduction.hpp:14,
                 from /home/jan/sycl-workspace/sycl/sycl/include/CL/sycl.hpp:16,
                 from /home/jan/sycl-workspace/sycl/sycl/source/backend/level_zero.cpp:9:
/home/jan/sycl-workspace/sycl/sycl/include/CL/sycl/detail/cstr.hpp:55:16: error: ‘__type_pack_element’ does not name a type
   55 |     using at = __type_pack_element<Idx, cstr<CharT, charpack>...>;
```

Do you require more info to investigate?

```
> g++ --version
g++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
Copyright (C) 2019 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

j-stephan:
Thanks! I'm able to build `sycl/unified/next` with gcc again.

PR #110, opened 2021-02-23
quick fix the current build

Ralender:
Merging all the changes from intel/sycl breaks lots of things but we can't keep having sycl/unifed/next not compiling.
so this is just a cherry-pick of 33a2868f77b86faa5b0d575789babc81867fe808 while i work on a proper merge.

### DIFF
--- a/buildbot/dependency.py
+++ b/buildbot/dependency.py
@@ -49,11 +49,17 @@ def do_dependency(args):
     # fetch OpenCL headers
     ocl_header_dir = os.path.join(args.obj_dir, "OpenCL-Headers")
     if not os.path.isdir(ocl_header_dir):
-        clone_cmd = ["git", "clone", "https://github.com/KhronosGroup/OpenCL-Headers", "OpenCL-Headers"]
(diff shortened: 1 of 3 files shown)


lforg37:
It seems that the issue #107 is back with this PR

lforg37 (approved):

keryell:
> It seems that the issue #107 is back with this PR

That seems unlikely since the C++ runtime has not been touched.

lforg37:
> That seems unlikely since the C++ runtime has not been touched.

Yes, I tested only the Ralender/FixMerge2 and not the merge with the main branch.
I deleted the original comment but I guess there is no edition delay before the mail notification is sent by github.
Sorry for the noise.

keryell:
> Yes, I tested only the Ralender/FixMerge2 and not the merge with the main branch. I deleted the original comment but I guess there is no edition delay before the mail notification is sent by github. Sorry for the noise.

It took me some time to track this comment to reply to. Since I was not able to find it in the PR, I reply to the e-mail I had received... :-)

PR #111, opened 2021-02-26
Merge upstream

Ralender:
This contains the merge from intel's sycl and fixes to sw_emu and hw_emu

### DIFF
--- a/clang/lib/AST/Expr.cpp
+++ b/clang/lib/AST/Expr.cpp
@@ -33,6 +33,7 @@
 #include "clang/Lex/LiteralSupport.h"
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/Support/SHA1.h"
 #include <algorithm>
 #include <cstring>
 using namespace clang;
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -15,14 +15,14 @@ opencl-aot/ @dm-vodopyanov @AlexeySachkov @romanovvlad
 # Device library
 libdevice/ @vzakhari
 
+# DPC++ runtime library
```
keryell, line 18: Perhaps we could start filling this part too... ;-)

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -30,6 +30,7 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
+#include "llvm/../../lib/IR/LLVMContextImpl.h"
```
keryell: This looks strange

clang/lib/Basic/Targets/SPIR.h
```diff
@@ -41,12 +41,12 @@ static const unsigned SPIRAddrSpaceMap[] = {
...
 
 static const unsigned SYCLAddrSpaceMap[] = {
-    4, // Default
+    0, // Default
     1, // opencl_global
     3, // opencl_local
     2, // opencl_constant
     0, // opencl_private
-    4, // opencl_generic
+    0, // opencl_generic
```
keryell: I have some doubts about having both 0 for `opencl_private` and `opencl_generic`.

Why does it works at the end? Actually does it work on real hardware? At least we should check that it fits what is consumed by Vitis through our massaging passes.

llvm/lib/Transforms/InstCombine/InstCombineLoadStoreAlloca.cpp
```diff
@@ -1447,6 +1457,15 @@ Instruction *InstCombinerImpl::visitStoreInst(StoreInst &SI) {
   if (isa<UndefValue>(Val))
     return eraseInstFromFunction(SI);
 
+  if (Triple(SI.getModule()->getTargetTriple()).isXilinxFPGA()) {
```
keryell: Add rationale as comments

llvm/lib/Transforms/InstCombine/InstCombineLoadStoreAlloca.cpp
```diff
@@ -1031,6 +1031,16 @@ Instruction *InstCombinerImpl::visitLoadInst(LoadInst &LI) {
         return replaceOperand(LI, 0, SI->getOperand(1));
     }
   }
+
+  if (Triple(LI.getModule()->getTargetTriple()).isXilinxFPGA()) {
```
keryell: Add rationale as comments

keryell (commented):
You know how I love comments... :-)

llvm/lib/SYCL/XOCCIRDowngrader.cpp
Ralender: yeah this is a dependency we rather not have. but using this is the most reliable way to eliminate poison.

clang/lib/Basic/Targets/SPIR.h
Ralender: I will need to do some more testing on this.
but this prevents a errors during compilation and seems to be working in sw_emu and hw_emu

llvm/lib/SYCL/XOCCIRDowngrader.cpp
keryell: Would `#include "../lib/IR/LLVMContextImpl.h"` work too and this is just to make clear this is related to `llvm`?

keryell: Ah actually you have already changed it!

Ralender:
I have seen a few issue. we should wait for me to sort them out before we merge this.

Ralender:
the issue was having no history of the merged commits but this is now fixed.

llvm/lib/Transforms/InstCombine/InstCombineLoadStoreAlloca.cpp
keryell: And looking again at it, would it not be clearer to have this into onr of our vinification (Vitis...) passes instead of scattered across some random passes, when possible?

keryell:
Thanks, the demo with `-fsycl-unnamed-lambda` worked today.
But I had the feeling this option was breaking the use case when I was giving a name like before.
But I need to make a deeper experimentation. It was just before my presentation...

PR #118, opened 2021-05-01
Update documentation to Vitis 2020.2

keryell:


sycl/doc/GettingStartedXilinxFPGA.md
```diff
@@ -278,7 +289,8 @@ The typical environment is setup with something like
 ```bash
 # The place where SYCL has been compiled:
 SYCL_HOME=~/sycl_workspace
-XILINX_VERSION=2020.1
+# The version of Vitis you want to use
+XILINX_VERSION=2020.3
```
Ralender: shoudn't we point users to 2020.2 instead ?

keryell: Yes. Actually I am working on another commit to remove all the 2020.3 references since it is irrelevant for the FPGA Alveo boards.

PR #119, opened 2021-05-05
Add test with rel_ptr

Ralender:
Make XOCCIRDowngrader warn about instruction not supported by the backend.
Make assert be noop on device.

### DIFF
--- a/llvm/lib/SYCL/XOCCIRDowngrader.cpp
+++ b/llvm/lib/SYCL/XOCCIRDowngrader.cpp
@@ -12,9 +12,12 @@
 // ===---------------------------------------------------------------------===//
 
 #include <cstddef>
+#include <functional>
 #include <regex>
 #include <string>
 
+#include "llvm/ADT/MapVector.h"
+#include "llvm/IR/InstVisitor.h"
 #include "llvm/IR/Instructions.h"
 #include "llvm/IR/IntrinsicInst.h"
 #include "llvm/IR/Intrinsics.h"
@@ -31,6 +34,10 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/Path.h"
 #include "llvm/Support/raw_ostream.h"
+
+/// This should theoretically not be accessed from outside of the IR directory.
+/// But using it is the most reliable way to do some of the IR transformation we
+/// are doing in this file.
 #include "llvm/../../lib/IR/LLVMContextImpl.h"
 
 using namespace llvm;
@@ -264,6 +271,67 @@ struct XOCCIRDowngrader : public ModulePass {
       I->eraseFromParent();
   }
 
+  /// This will remove every call to the function named Str assuming it returns
+  /// void. and erase the function form the module.
+  void removeFunction(Module &M, StringRef Str) {
+    Function *F = M.getFunction(Str);
+    if (!F)
+      return;
+    SmallVector<std::reference_wrapper<Use>, 16> ToDelete;
+    ToDelete.append(F->use_begin(), F->use_end());
+    for (Use &U : ToDelete) {
+      assert(U.getUser()->use_empty() &&
+             "this should only be used on functions returning void");
+      if (auto *I = dyn_cast<Instruction>(U.getUser()))
+        I->eraseFromParent();
+      else
+        U.set(UndefValue::get(U->getType()));
+    }
+    F->eraseFromParent();
+  }
+
+  /// Visit the IR and emit warnings about construct not handled by the backend
+  /// The IR has no debug info so we cannot say where in the source code the
+  /// error happend.
+  struct WarnVisitor : InstVisitor<WarnVisitor> {
+    /// This is used for dedupping warnings.
+    MapVector<std::string, int, StringMap<int>> DiagMap;
+    /// Add a warning to be emmitted
+    template <typename T, typename... Ts> void warn(T P, Ts... Ps) {
+      std::string str;
+      raw_string_ostream os(str);
+      os << P;
+      (void)std::initializer_list<int>{(os << Ps, 0)...};
+      DiagMap[str]++;
+    }
+    /// This will be called by the visitor on every instruction in the module.
+    void visitInstruction(Instruction& I) {
+      switch (I.getOpcode()) {
+      case Instruction::IntToPtr:
+      case Instruction::PtrToInt:
+      case Instruction::AddrSpaceCast:
+        warn("instruction not supported by backend: \"", I.getOpcodeName(), "\"");
+      }
+    }
+    /// This will do the actual printing of warnings to the console.
+    void emit() {
+      if (DiagMap.empty())
+        return;
+      llvm::errs() << "\n";
+      for (auto &Elem : DiagMap)
+        llvm::errs() << raw_ostream::MAGENTA << "warning:" << raw_ostream::RESET
+                     << " " << Elem.first << " : " << Elem.second
+                     << " occurrences\n";
+      llvm::errs() << "\n";
+    }
+  };
+
+  void warnForIssues(Module &M) {
+    WarnVisitor Visitor;
+    Visitor.visit(M);
+    Visitor.emit();
+  }
+
   bool runOnModule(Module &M) override {
     resetByVal(M);
     removeAttributes(M, {Attribute::WillReturn, Attribute::NoFree,
@@ -276,6 +344,9 @@ struct XOCCIRDowngrader : public ModulePass {
 
     lowerIntrinsic(M);
     removeMetaDataValues(M);
+    /// __assert_fail doesn't exist on device and takes its arguments in
+    /// addressspace 0 causing addresspace cast.
+    removeFunction(M, "__assert_fail");
 
     convertPoinsonToZero(M);
     if (Triple(M.getTargetTriple()).getArch() == llvm::Triple::fpga64)
@@ -283,6 +354,9 @@ struct XOCCIRDowngrader : public ModulePass {
     else
       M.setTargetTriple("fpga32-xilinx-none");
     // The module probably changed
+
+    warnForIssues(M);
+
     return true;
   }
 };
--- a/llvm/lib/SYCL/XOCCIRDowngrader.cpp
+++ b/llvm/lib/SYCL/XOCCIRDowngrader.cpp
@@ -272,7 +272,7 @@ struct XOCCIRDowngrader : public ModulePass {
   }
 
   /// This will remove every call to the function named Str assuming it returns
-  /// void. and erase the function form the module.
+  /// void. Also erase the function from the module.
   void removeFunction(Module &M, StringRef Str) {
     Function *F = M.getFunction(Str);
     if (!F)
@@ -326,10 +326,30 @@ struct XOCCIRDowngrader : public ModulePass {
     }
   };
 
+  /// Traverse the IR in the module and warn about IR constructs unsupported by
+  /// the backend.
   void warnForIssues(Module &M) {
     WarnVisitor Visitor;
     Visitor.visit(M);
     Visitor.emit();
+  }
+
+  void cleanSpirBuiltins(Module &M) {
+    /// Find function
+    auto *spirid = M.getFunction("llvm.spir.get.global.id.64");
+    /// Create replcement
+    auto *replacement = ConstantInt::get(spirid->getReturnType(), 0);
+    if (spirid != nullptr && spirid->isDeclaration())
+      for (auto *user : spirid->users())
+        /// find all users
+        if (auto *call = dyn_cast<CallBase>(user)) {
+          /// replace calls by constant
+          call->replaceAllUsesWith(replacement);
+          call->eraseFromParent();
+        }
+    assert(spirid->use_empty());
+    /// erase the fontion from the module.
+    spirid->eraseFromParent();
   }
 
   bool runOnModule(Module &M) override {
--- a/llvm/lib/SYCL/XOCCIRDowngrader.cpp
+++ b/llvm/lib/SYCL/XOCCIRDowngrader.cpp
@@ -337,18 +337,18 @@ struct XOCCIRDowngrader : public ModulePass {
   void cleanSpirBuiltins(Module &M) {
     /// Find function
     auto *spirid = M.getFunction("llvm.spir.get.global.id.64");
-    /// Create replcement
+    /// Create replacement
     auto *replacement = ConstantInt::get(spirid->getReturnType(), 0);
     if (spirid != nullptr && spirid->isDeclaration())
       for (auto *user : spirid->users())
         /// find all users
         if (auto *call = dyn_cast<CallBase>(user)) {
-          /// replace calls by constant
+          /// Replace calls by constant
           call->replaceAllUsesWith(replacement);
           call->eraseFromParent();
         }
     assert(spirid->use_empty());
-    /// erase the fontion from the module.
+    /// Erase the function from the module.
     spirid->eraseFromParent();
   }
 
--- a/llvm/lib/SYCL/XOCCIRDowngrader.cpp
+++ b/llvm/lib/SYCL/XOCCIRDowngrader.cpp
@@ -334,24 +334,6 @@ struct XOCCIRDowngrader : public ModulePass {
     Visitor.emit();
   }
 
-  void cleanSpirBuiltins(Module &M) {
-    /// Find function
-    auto *spirid = M.getFunction("llvm.spir.get.global.id.64");
-    /// Create replacement
-    auto *replacement = ConstantInt::get(spirid->getReturnType(), 0);
-    if (spirid != nullptr && spirid->isDeclaration())
-      for (auto *user : spirid->users())
-        /// find all users
-        if (auto *call = dyn_cast<CallBase>(user)) {
-          /// Replace calls by constant
-          call->replaceAllUsesWith(replacement);
-          call->eraseFromParent();
-        }
-    assert(spirid->use_empty());
-    /// Erase the function from the module.
-    spirid->eraseFromParent();
-  }
-
   bool runOnModule(Module &M) override {
     resetByVal(M);
     removeAttributes(M, {Attribute::WillReturn, Attribute::NoFree,
--- a/sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
+++ b/sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
@@ -1,0 +1,235 @@
+// REQUIRES: xocc
+
+// RUN: %clangxx -fsycl -fsycl-targets=%sycl_triple -fsycl-unnamed-lambda -std=c++20 -Xsycl-target-frontend -fno-exceptions %s -o %t.out
+// RUN: %ACC_RUN_PLACEHOLDER %t.out
+
+#include <CL/sycl.hpp>
+#include <cassert>
+#include <optional>
+#include <type_traits>
+#include <variant>
+namespace detail {
+
+template <typename> struct variant_trait {};
+
+template <typename... Tys> struct variant_trait<std::variant<Tys...>> {
+  using indexes = std::make_index_sequence<sizeof...(Tys)>;
+};
+
+// // This could be a replacement for visit_single_impl but causes addrespace casts
+// template <typename ret_type, typename Func, typename Var, auto... Idx>
+// inline ret_type
+// visit_single_impl(Func &&f, std::integer_sequence<size_t, Idx...>, Var &&var) {
+//   if constexpr (std::is_same<void, ret_type>::value) {
+//     (
+//         [&] {
+//           if (var.index() == Idx)
+//             std::forward<Func>(f)(std::get<Idx>(var));
+//         }(),
+//         ...);
+//   } else {
+//     // we use optional here because ret_type may not
+//     // be default constructible.
+//     std::optional<ret_type> ret;
+//     (
+//         [&] {
+//           if (var.index() == Idx)
+//             ret.emplace(std::forward<Func>(f)(std::get<Idx>(var)));
+//         }(),
+//         ...);
+//     return std::forward<ret_type>(*ret);
+//   }
+// }
+
+template <typename ret_type, typename Func, typename Var>
+[[noreturn]] inline ret_type
+visit_single_impl(Func&&, std::integer_sequence<size_t>, Var&&) {
+  // assert(false && "unreachable");
+  __builtin_unreachable();
+}
+
+template <typename ret_type, typename Func, typename Var, auto First,
+          auto... Idx>
+inline ret_type visit_single_impl(Func&& f,
+                                  std::integer_sequence<size_t, First, Idx...>,
+                                  Var&& var) {
+  if (var.index() == First)
+    return std::forward<Func>(f)(std::get<First>(var));
+  return visit_single_impl<ret_type>(std::forward<Func>(f),
+                                     std::integer_sequence<size_t, Idx...> {},
+                                     std::forward<Var>(var));
+}
+
+template <typename Func, typename Var>
+decltype(auto) visit_single(Func&& f, Var&& var) {
+  assert((!var.valueless_by_exception()));
+  using ret_type =
+      std::invoke_result_t<Func, decltype(std::get<0>(std::declval<Var>()))>;
+  return visit_single_impl<ret_type>(
+      std::forward<Func>(f),
+      typename variant_trait<
+          std::remove_cv_t<std::remove_reference_t<Var>>>::indexes {},
+      std::forward<Var>(var));
+}
+} // namespace detail
+
+/// dev_visit is std::visit implementation suitable to be used in device code.
+/// this version of visit doesn't use any function pointer but uses if series
+/// which will be turned into switch case by the optimizer.
+template <typename Func, typename Var, typename... Rest>
+auto dev_visit(Func&& f, Var&& var, Rest&&... rest) {
+  if constexpr (sizeof...(Rest) == 0)
+    return detail::visit_single(std::forward<Func>(f), std::forward<Var>(var));
+  else
+    return detail::visit_single(
+        [&](auto&& First) {
+          return dev_visit(
+              [&](auto&&... Others) {
+                std::forward<Func>(f)(
+                    std::forward<decltype(First)>(First),
+                    std::forward<decltype(Others)>(Others)...);
+              },
+              std::forward<Rest>(rest)...);
+        },
+        std::forward<Var>(var));
+}
+
+
+/// This is a relative pointer behaving mostly like T*. It doesn't manage the
+/// lifetime of the data pointed to. Also this class have shalow const semantic
+/// meaning that const rel_ptr<T> is equivalent to T* const not const T *. This
+/// pointer should behave like T* for any operation except memcpy or bitcasting
+/// the representation of the pointer. When memcpy'ed to an other place the
+/// offset between the pointer's address and the data it is pointed to will stay
+/// constant. this means that if the data is copied with the same memcpy the
+/// pointer is still pointing to the memcpy'ed data.
+/// This pointer can also be used to store pointers into files.
+template <typename T, typename IntTy = std::ptrdiff_t> class rel_ptr {
+  IntTy offset;
+
+  char *get_this() const {
+    return reinterpret_cast<char *>(const_cast<rel_ptr *>(this));
+  }
+  IntTy get_offset(T *t) const {
+    return (reinterpret_cast<char *>(t) - get_this());
+  }
+
+public:
+  T *get() const { return (T *)(get_this() + offset); }
+  T &operator*() const { return *get(); }
+  T *operator->() const { return get(); }
+  T &operator[](std::size_t e) const { return get()[e]; }
+  explicit operator bool() const { return get(); }
+  bool operator!() const { return !get(); }
+
+  rel_ptr() : rel_ptr(nullptr) {}
+  explicit rel_ptr(T *t) : offset(get_offset(t)) {}
+  rel_ptr(const rel_ptr &other) : rel_ptr(other.get()) {}
+  rel_ptr &operator=(const rel_ptr &other) {
+    offset = get_offset(other.get());
+    return *this;
+  }
+
+  friend rel_ptr operator+(const rel_ptr &ptr, IntTy off) {
+    return rel_ptr(ptr.get() + off);
+  }
+  friend rel_ptr operator-(const rel_ptr &ptr, IntTy off) {
+    return rel_ptr(ptr.get() - off);
+  }
+  IntTy operator-(const rel_ptr &other) const { return get() - other.get(); }
+
+  bool operator==(const rel_ptr &other) const { return get() == other.get(); }
+  friend bool operator==(const rel_ptr &p1, T *p2) { return p1.get() == p2; }
+
+  /// we use weak_ordering because the bitcasted representation of equal pointer
+  /// is different. because 2 pointers are equal if they point to the same
+  /// object and in order to pointe to the same object they have different
+  /// offsets to it so there representation is different.
+  std::weak_ordering operator<=>(const rel_ptr &other) {
+    return get() <=> other.get();
+  }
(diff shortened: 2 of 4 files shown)


llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -14,7 +14,10 @@
 #include <cstddef>
 #include <regex>
 #include <string>
+#include <functional>
```
keryell: Sort

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+  sycl::buffer<node> In{data.data(), {data.size()}};
+  sycl::buffer<int> Out{1};
+
+  sycl::queue Queue{};
+
+  Queue.submit([&](sycl::handler &cgh) {
+    auto AIn = In.get_access<sycl::access::mode::read_write>(cgh);
+    auto AOut = Out.get_access<sycl::access::mode::write>(cgh);
+    int root_node = data.size() - 1;
+    cgh.single_task<class Kernel>([=]() {
```
keryell:
```suggestion
    cgh.single_task<class Kernel>([=] {
```

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +267,58 @@ struct XOCCIRDowngrader : public ModulePass {
       I->eraseFromParent();
   }
 
+  void removeFunction(Module &M, StringRef Str) {
```
keryell, line 276: Curious this is not already in LLVM utils.
Add a comment describing what the function does.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +267,58 @@ struct XOCCIRDowngrader : public ModulePass {
...
+             "this should only be used on functions returning void");
+      if (auto *I = dyn_cast<Instruction>(U.getUser()))
+        I->eraseFromParent();
+      else
+        U.set(UndefValue::get(U->getType()));
+    }
+    F->eraseFromParent();
+  }
+
+  struct WarnVisitor : InstVisitor<WarnVisitor> {
```
keryell, line 296: Add a comment describing what the function does.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +267,58 @@ struct XOCCIRDowngrader : public ModulePass {
...
+    MapVector<std::string, int, StringMap<int>> DiagMap;
+    template <typename T, typename... Ts> void warn(T P, Ts... Ps) {
+      std::string str;
+      raw_string_ostream os(str);
+      os << P;
+      (void)std::initializer_list<int>{(os << Ps, 0)...};
+      DiagMap[str]++;
+    }
+    void visitInstruction(Instruction& I) {
```
keryell, line 308: Add a comment describing what the function does.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +267,58 @@ struct XOCCIRDowngrader : public ModulePass {
...
+      switch (I.getOpcode()) {
+      case Instruction::IntToPtr:
+      case Instruction::PtrToInt:
+      case Instruction::AddrSpaceCast:
+        warn("instruction not supported by backend: \"", I.getOpcodeName(), "\"");
+      }
+    }
+    void emit() {
```
keryell, line 317: Add a comment describing what the function does.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +267,58 @@ struct XOCCIRDowngrader : public ModulePass {
...
+      llvm::errs() << "\n";
+      for (auto &Elem : DiagMap)
+        llvm::errs() << raw_ostream::MAGENTA << "warning:" << raw_ostream::RESET
+                     << " " << Elem.first << " : " << Elem.second
+                     << " occurrences\n";
+      llvm::errs() << "\n";
+    }
+  };
+
+  void warnForIssues(Module &M) {
```
keryell, line 331: Add a comment describing what the function does.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +267,58 @@ struct XOCCIRDowngrader : public ModulePass {
...
+      os << P;
+      (void)std::initializer_list<int>{(os << Ps, 0)...};
+      DiagMap[str]++;
+    }
+    void visitInstruction(Instruction& I) {
+      switch (I.getOpcode()) {
+      case Instruction::IntToPtr:
+      case Instruction::PtrToInt:
+      case Instruction::AddrSpaceCast:
+        warn("instruction not supported by backend: \"", I.getOpcodeName(), "\"");
```
keryell, line 313: It might be not enough to just know the instruction to fix the problem.
What about dumping the source-location if we have it?

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -276,13 +331,19 @@ struct XOCCIRDowngrader : public ModulePass {
 
     lowerIntrinsic(M);
     removeMetaDataValues(M);
+    /// __assert_fail doesn't exist on device and takes its arguments in
+    /// addressspace 0 causing addresspace cast.
```
keryell, line 350: I appreciate the comment. :-)

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -276,13 +331,19 @@ struct XOCCIRDowngrader : public ModulePass {
...
+    removeFunction(M, "__assert_fail");
 
     convertPoinsonToZero(M);
     if (Triple(M.getTargetTriple()).getArch() == llvm::Triple::fpga64)
       M.setTargetTriple("fpga64-xilinx-none");
     else
       M.setTargetTriple("fpga32-xilinx-none");
     // The module probably changed
```
keryell, line 358: This comment goes with the `return true;`.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+template <typename> struct variant_trait {};
+
+template <typename... Tys> struct variant_trait<std::variant<Tys...>> {
+  using indexes = std::make_index_sequence<sizeof...(Tys)>;
+};
+
+template <typename ret_type, typename Func, typename Var>
+[[noreturn]] inline ret_type
+visit_single_impl(Func&&, std::integer_sequence<size_t>, Var&&) {
+  // assert(false && "unreachable");
```
keryell: I guess you can now use the `assert()`?

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+/// This is a relative pointer behaving mostly like T*. It doesn't manage the
+/// lifetime of the data pointed to. Also this class have shalow const semantic
+/// meaning that const rel_ptr<T> is equivalent to T* const not const T *. This
+/// pointer should behave like T* for any operation except memcpy or bitcasting
+/// the representation of the pointer. When memcpy'ed to an other place the
+/// offset between the pointer's address and the data it is pointed to will stay
+/// constant. this means that if the data is copied with the same memcpy the
+/// pointer is still pointing to the memcpy'ed data.
+/// This pointer can also be used to store pointers into files.
+template <typename T, typename IntTy = int64_t> class rel_ptr {
```
keryell: `template <typename T, typename IntTy = std::intptr_t> class rel_ptr {`

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+  IntTy offset;
+
+  char *get_this() const { return ((char *)this); }
```
keryell: Replace all the C casts by C++ casts.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+  IntTy get_offset(T *t) const { return (((char *)t) - get_this()); }
+
+public:
+  T *get() const { return (T *)(get_this() + offset); }
+  T &operator*() const { return *get(); }
+  T *operator->() const { return get(); }
+  T &operator[](size_t e) const { return get()[e]; }
```
keryell: `std::size_t`

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+  rel_ptr() : rel_ptr(nullptr) {}
+  rel_ptr(T *t) : offset(get_offset(t)) {}
+  rel_ptr(const rel_ptr &other) : rel_ptr(other.get()) {}
+  rel_ptr &operator=(const rel_ptr &other) {
+    offset = get_offset(other.get());
+    return *this;
+  }
+
+  rel_ptr operator+(IntTy off) { return rel_ptr(get() + off); }
+  rel_ptr operator-(IntTy off) { return rel_ptr(get() - off); }
```
keryell: Use hidden friends to handle `5 + p`

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+  IntTy operator-(const rel_ptr &other) const { return offset - other.offset; }
+
+  bool operator==(const rel_ptr &other) const { return get() == other.get(); }
+  bool operator!=(const rel_ptr &other) const { return !(*this == other); }
+  friend bool operator==(const rel_ptr &p1, T *p2) { return p1.get() == p2; }
+  friend bool operator!=(const rel_ptr &p1, T *p2) { return !(p1 == p2); }
```
keryell: Just use 
```c++
friend auto operator<=>(const rel_ptr&, const rel_ptr&) = default;
```
and rely on implicit conversion from plain pointer.
Add a few comments all over the place to explain what's happening in this code.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+} // namespace detail
+
+/// dev_visit is std::visit implementation suitable to be used in device code.
+/// this version of visit doesn't use any function pointer but uses if series
+/// which will be turned into switch case by the optimizer.
+template <typename Func, typename Var, typename... Rest>
+auto dev_visit(Func&& f, Var&& var, Rest&&... rest) {
+  if constexpr (sizeof...(Rest) == 0)
+    return detail::visit_single(std::forward<Func>(f), std::forward<Var>(var));
```
keryell, line 83: I have the feeling that the recursion on `dev_visit` could be replaced by some https://www.foonathan.net/2020/05/fold-tricks/

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+int main() {
+  // computation graph for (1 + 2) * 3
+  std::array<node, 5> data{{Constant{1}, Constant{2}, AddOp{&data[0], &data[1]},
+                            Constant{3}, MulOp{&data[2], &data[3]}}};
+  sycl::buffer<node> In{data.data(), {data.size()}};
+  sycl::buffer<int> Out{1};
+
+  sycl::queue Queue{};
```
keryell: `sycl::queue Queue;`

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+struct node;
+
+struct BinOp { rel_ptr<node> elem[2]; };
+struct AddOp : BinOp { int compute(); };
+struct MulOp : BinOp { int compute(); };
+
+struct node {
+  std::variant<Constant, AddOp, MulOp> n;
+  int cache;
```
keryell, line 195: The implentation is somehow broken since you do not really use a cache and so only works with a topologically sorted DAG.
So why not really have `std::optional<int> cache` instead and move `compute()` in `get_value()` instead?
But here we need to remove a recursion...
So perhaps a more generic way would be with no recursion to just return the optional with a value only if the value is there or compute the value if the operands have already a value... Cf infra

keryell: Actually implementing `compute()` doing the computation if not done already is still useful.

keryell (changes requested):
Quite interesting.
This looks like a great coding interview for modern C++ by the way. :-)

keryell:
Also the evaluation_tree.cpp could be named DAG evaluation.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
Ralender: This isn't in LLVM because it remove function even if they have uses. which is generally not done in LLVM because it usually changes the behavior of the program

llvm/lib/SYCL/XOCCIRDowngrader.cpp
Ralender: source-locations aren't available because our backend doesn't support them.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
Ralender: i used `std::ptrdiff_t` instead.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
Ralender: personally find C casts more readable in this kind of situations. but I changed it.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
Ralender:
```
friend auto operator<=>(const rel_ptr&, const rel_ptr&) = default;
```
is wrong since it will compare `offset` instead of the object pointed to (`get()`)
letting implicit conversion from T* to  rel_ptr<T> was an oversight.

however I used operator<=>.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
Ralender: using compute form get_value is not possible.
because the backend compiler requires no recursion in the IR. and there would be a cicle like: `get_value -> compute -> AddOp::compute -> get_value`

but a agree that it is not such a good example without the recurtion because it requires a topologically sorted DAG and is generally not nice looking code. nevertheless rel_ptr made it look much better than before.

Ralender:
This is now based on https://github.com/triSYCL/sycl/pull/120.
I had some issues with addressspacecast because of the changes to dev_visit so commented the changes and reverted them.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
keryell: Clarify the comment above, then.

keryell:
Why are there so many commits if it is based on https://github.com/triSYCL/sycl/pull/120 ?

Ralender:
> Why are there so many commits if it is based on #120 ?

This is because the merge commit of sycl/unified/next that wasn't in this PR because it didn't exist at the time.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +271,67 @@ struct XOCCIRDowngrader : public ModulePass {
       I->eraseFromParent();
   }
 
+  /// This will remove every call to the function named Str assuming it returns
+  /// void. and erase the function form the module.
```
keryell: `/// void. Also erase the function from the module.`

llvm/lib/SYCL/XOCCIRDowngrader.cpp
keryell: Sure, but some day...

llvm/lib/SYCL/XOCCIRDowngrader.cpp
keryell: You can get inspiration from the doc on the visitor, just a little bit more high-level here.

llvm/lib/SYCL/XOCCIRDowngrader.cpp
keryell: Is this the real assert implementation or the extension from Intel?

keryell:
Ping

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +271,87 @@ struct XOCCIRDowngrader : public ModulePass {
...
+    MapVector<std::string, int, StringMap<int>> DiagMap;
+    /// Add a warning to be emmitted
+    template <typename T, typename... Ts> void warn(T P, Ts... Ps) {
+      std::string str;
+      raw_string_ostream os(str);
+      os << P;
+      (void)std::initializer_list<int>{(os << Ps, 0)...};
+      DiagMap[str]++;
+    }
+    /// This will be called by the visitor on every instruction in the module.
```
keryell, line 307: Add a new line before probably

llvm/lib/SYCL/XOCCIRDowngrader.cpp
keryell: But also add a new line before...

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +271,87 @@ struct XOCCIRDowngrader : public ModulePass {
...
+  void warnForIssues(Module &M) {
+    WarnVisitor Visitor;
+    Visitor.visit(M);
+    Visitor.emit();
+  }
+
+  void cleanSpirBuiltins(Module &M) {
+    /// Find function
+    auto *spirid = M.getFunction("llvm.spir.get.global.id.64");
+    /// Create replcement
```
keryell:
```suggestion
    /// Create replacement
```

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +271,87 @@ struct XOCCIRDowngrader : public ModulePass {
...
+    auto *replacement = ConstantInt::get(spirid->getReturnType(), 0);
+    if (spirid != nullptr && spirid->isDeclaration())
+      for (auto *user : spirid->users())
+        /// find all users
+        if (auto *call = dyn_cast<CallBase>(user)) {
+          /// replace calls by constant
```
keryell:
```suggestion
          /// Replace calls by constant
```

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +271,87 @@ struct XOCCIRDowngrader : public ModulePass {
...
+          call->replaceAllUsesWith(replacement);
+          call->eraseFromParent();
+        }
+    assert(spirid->use_empty());
+    /// erase the fontion from the module.
```
keryell:
```suggestion
    /// Erase the function from the module.
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include <variant>
+namespace detail {
+
+template <typename> struct variant_trait {};
+
+template <typename... Tys> struct variant_trait<std::variant<Tys...>> {
+  using indexes = std::make_index_sequence<sizeof...(Tys)>;
+};
+
+// // This could be a replacement for visit_single_impl but causes addrespace casts
```
keryell:
```suggestion
// // This could be a replacement for visit_single_impl but causes address space casts
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
keryell: Add a comment instead explaining `assert` is not usable on device so use `__builtin_unreachable()` instead?

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+  return visit_single_impl<ret_type>(
+      std::forward<Func>(f),
+      typename variant_trait<
+          std::remove_cv_t<std::remove_reference_t<Var>>>::indexes {},
+      std::forward<Var>(var));
+}
+} // namespace detail
+
+/// dev_visit is std::visit implementation suitable to be used in device code.
+/// this version of visit doesn't use any function pointer but uses if series
```
keryell:
```suggestion
/// This version of visit doesn't use any function pointer but uses if series
```
I do not understand what you mean by "if series" and the following,

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+                    std::forward<decltype(Others)>(Others)...);
+              },
+              std::forward<Rest>(rest)...);
+        },
+        std::forward<Var>(var));
+}
+
+
+/// This is a relative pointer behaving mostly like T*. It doesn't manage the
+/// lifetime of the data pointed to. Also this class have shalow const semantic
```
keryell:
```suggestion
/// lifetime of the data pointed to. Also this class have shallow const semantic
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+/// meaning that const rel_ptr<T> is equivalent to T* const not const T *. This
+/// pointer should behave like T* for any operation except memcpy or bitcasting
+/// the representation of the pointer. When memcpy'ed to an other place the
```
keryell:
```suggestion
/// the representation of the pointer. When memcpy'ed to another place the
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+/// offset between the pointer's address and the data it is pointed to will stay
+/// constant. this means that if the data is copied with the same memcpy the
```
keryell:
```suggestion
/// constant. This means that if the data is copied with the same memcpy the
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+/// pointer is still pointing to the memcpy'ed data.
+/// This pointer can also be used to store pointers into files.
```
keryell: Unclear how it can be done and what it means, which hypothesis?

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+template <typename T, typename IntTy = std::ptrdiff_t> class rel_ptr {
+  IntTy offset;
+
+  char *get_this() const {
```
keryell:
```suggestion
  char* get_this() const {
```
and so on to be consistent with previous code

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+    return reinterpret_cast<char *>(const_cast<rel_ptr *>(this));
```
keryell:
```suggestion
    return reinterpret_cast<char*>(const_cast<rel_ptr*>(this));
```
and so on

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+  }
+  friend rel_ptr operator-(const rel_ptr &ptr, IntTy off) {
+    return rel_ptr(ptr.get() - off);
+  }
+  IntTy operator-(const rel_ptr &other) const { return get() - other.get(); }
+
+  bool operator==(const rel_ptr &other) const { return get() == other.get(); }
+  friend bool operator==(const rel_ptr &p1, T *p2) { return p1.get() == p2; }
+
+  /// we use weak_ordering because the bitcasted representation of equal pointer
```
keryell:
```suggestion
  /// We use weak_ordering because the bitcasted representation of equal pointer
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+  /// is different. because 2 pointers are equal if they point to the same
```
keryell:
```suggestion
  /// is different. 2 pointers are equal if they point to the same
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+  /// object and in order to pointe to the same object they have different
```
keryell:
```suggestion
  /// object and in order to point to the same object they have different
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,235 @@
...
+  /// offsets to it so there representation is different.
```
keryell:
```suggestion
  /// offsets to it, so their representations are different.
```

keryell (changes requested):
Making progress.
But find a way to have a spellchecker working in your IDE! :-(

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
Ralender: storing data to files by reinterpreting it in bytes and reading it by reinterpreting bytes into structed data is kind of equivalent to a memcpy across address space.  if you simply mmap the file into memory you can use rel_ptr to follow references between different parts of the file.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,240 @@
...
+        std::forward<Var>(var));
+}
+
+/// This is a relative pointer behaving mostly like T*. It doesn't manage the
+/// lifetime of the data pointed to. Also this class have shallow const semantic
+/// meaning that const rel_ptr<T> is equivalent to T* const not const T *. This
+/// pointer should behave like T* for any operation except memcpy or bitcasting
+/// the representation of the pointer. When memcpy'ed to another place the
+/// offset between the pointer's address and the data it is pointed to will stay
+/// constant. This means that if the data is copied with the same memcpy the
```
keryell:
```suggestion
/// constant. This means that, if a pointer and its pointed data are copied with the same memcpy, the
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,240 @@
...
+/// pointer is still pointing to the same data.
+/// storing data to files by reinterpreting it in bytes and reading it by
+/// reinterpreting bytes into structured data is kind of equivalent to a memcpy
```
keryell:
```suggestion
/// Storing data to files by reinterpreting it in bytes and reading it by
/// reinterpreting bytes into structured data is kind of equivalent to a memcpy
```
but I do not understand what you mean here. Let's talk about it on Tuesday meeting.

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,240 @@
...
+/// across address space. if you simply mmap the file into memory you can use
+/// rel_ptr to follow references between different parts of the file. This
+/// pointer can also be used to store references between various parts of the
+/// same file.
```
keryell: It looks like the last sentence does not add much to the previous sentence. Is there a way just to improve the previous sentence?

keryell (changes requested):

llvm/lib/SYCL/XOCCIRDowngrader.cpp
```diff
@@ -264,6 +271,87 @@ struct XOCCIRDowngrader : public ModulePass {
...
+  }
+
+  void cleanSpirBuiltins(Module &M) {
+    /// Find function
+    auto *spirid = M.getFunction("llvm.spir.get.global.id.64");
+    /// Create replacement
+    auto *replacement = ConstantInt::get(spirid->getReturnType(), 0);
+    if (spirid != nullptr && spirid->isDeclaration())
+      for (auto *user : spirid->users())
+        /// find all users
```
keryell:
```suggestion
        /// Find all users
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,238 @@
...
+int AddOp::compute() {
+  return BinOp::elem[0]->get_value() + BinOp::elem[1]->get_value();
+}
+
+int MulOp::compute() {
+  return BinOp::elem[0]->get_value() * BinOp::elem[1]->get_value();
+}
+
+int main() {
+  // computation graph for (1 + 2) * 3
```
keryell:
```suggestion
  // Computation graph for (1 + 2) * 3
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,238 @@
...
+/// lifetime of the data pointed to. Also this class have shallow const semantic
+/// meaning that const rel_ptr<T> is equivalent to T* const not const T *. This
+/// pointer should behave like T* for any operation except memcpy or bitcasting
+/// the representation of the pointer. When memcpy'ed to another place the
+/// offset between the pointer's address and the data it is pointed to will stay
+/// constant. This means that, if a pointer and its pointed data are copied with
+/// the same memcpy, the pointer is still pointing to the same data. storing
+/// data to files by reinterpreting it in bytes and reading it by reinterpreting
+/// bytes into structured data is kind of equivalent to a memcpy across address
+/// space. if you simply mmap the file into memory you can use rel_ptr to follow
```
keryell:
```suggestion
/// space. If you simply mmap the file into memory you can use rel_ptr to follow
```

sycl/test/on-device/xocc/simple_tests/evaluation_tree.cpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+/// This is a relative pointer behaving mostly like T*. It doesn't manage the
+/// lifetime of the data pointed to. Also this class have shallow const semantic
+/// meaning that const rel_ptr<T> is equivalent to T* const not const T *. This
+/// pointer should behave like T* for any operation except memcpy or bitcasting
+/// the representation of the pointer. When memcpy'ed to another place the
+/// offset between the pointer's address and the data it is pointed to will stay
+/// constant. This means that, if a pointer and its pointed data are copied with
+/// the same memcpy, the pointer is still pointing to the same data. storing
```
keryell:
```suggestion
/// the same memcpy, the pointer is still pointing to the same data. Storing
```

keryell (changes requested):
Almost there.
One of the change was already suggested previously...

keryell:
Great!

PR #147, opened 2021-10-23
Add generic_executor test from triSYCL

keryell:
Example for CppCon 2021 presentation.
Modernized version of
https://github.com/triSYCL/triSYCL/blob/master/tests/examples/generic_executor.cpp


keryell:
This works on HLS `hw_emu` flow on my laptop with current `triSYCL:sycl/unified/next` branch.

PR #153, opened 2021-12-06
[SYCL][FPGA][ACAP] Backport some downstream changes.

Ralender:
Factorize some functionalities such that they can be shared across SYCL passes

### DIFF
--- a/clang/lib/Sema/SemaSYCL.cpp
+++ b/clang/lib/Sema/SemaSYCL.cpp
@@ -34,6 +34,8 @@
 #include <functional>
 #include <initializer_list>
 
+#define DEBUG_TYPE "sema-sycl"
+
 using namespace clang;
 using namespace std::placeholders;
 
@@ -765,6 +767,14 @@ class SingleDeviceFunctionTracker {
     // If this isn't a function, I don't think there is anything we can do here.
     if (!CurrentDecl)
       return;
+
+    /// This will print a uniqued call graph of functions being complied for the
+    /// device. This is useful for debugging how a function ended up in device
+    /// code.
+    LLVM_DEBUG(if (!DeviceFunctions.contains(CurrentDecl)) {
+      llvm::dbgs() << std::string(CallStack.size(), ' ');
+      llvm::dbgs() << CurrentDecl->getQualifiedNameAsString() << "\n";
+    });
 
     // Determine if this is a recursive function. If so, we're done.
     if (llvm::is_contained(CallStack, CurrentDecl)) {
--- a/llvm/lib/SYCL/PrepareSYCLOpt.cpp
+++ b/llvm/lib/SYCL/PrepareSYCLOpt.cpp
@@ -29,6 +29,8 @@
 #include "llvm/Support/CommandLine.h"
 #include "llvm/Transforms/Utils/Cloning.h"
 
+#include "SYCLUtils.h"
+
 using namespace llvm;
 
 namespace {
@@ -81,7 +83,7 @@ struct PrepareSYCLOpt : public ModulePass {
         continue;
       }
       // Annotate kernels for HLS backend being able to identify them
-      if (F.getCallingConv() == CallingConv::SPIR_KERNEL) {
+      if (isKernelFunc(&F)) {
         assert(F.use_empty());
         F.addFnAttr("fpga.top.func", F.getName());
         F.addFnAttr("fpga.demangled.name", F.getName());
@@ -102,7 +104,7 @@ struct PrepareSYCLOpt : public ModulePass {
 
   void setCallingConventions(Module &M) {
     for (Function &F : M.functions()) {
-      if (F.getCallingConv() == CallingConv::SPIR_KERNEL) {
+      if (isKernelFunc(&F)) {
         assert(F.use_empty());
         continue;
       }
@@ -155,7 +157,7 @@ struct PrepareSYCLOpt : public ModulePass {
 
   void forceInlining(Module &M) {
     for (auto &F : M.functions()) {
-      if (F.isDeclaration() || F.getCallingConv() == CallingConv::SPIR_KERNEL)
+      if (F.isDeclaration() || isKernelFunc(&F))
         continue;
       F.addFnAttr(Attribute::AlwaysInline);
     }
@@ -220,7 +222,7 @@ struct PrepareSYCLOpt : public ModulePass {
   void unwrapFPGAProperties(Module &M) {
     UnwrapperVisitor UWV{};
     for (auto &F : M.functions()) {
-      if (F.getCallingConv() == CallingConv::SPIR_KERNEL) {
+      if (isKernelFunc(&F)) {
         UWV.visit(F);
       }
     }
--- a/llvm/lib/SYCL/SYCLUtils.cpp
+++ b/llvm/lib/SYCL/SYCLUtils.cpp
@@ -1,0 +1,69 @@
+//===- SYCLUtils.cpp ------------------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+//
+// Shared utilities between the various SYCL passes.
+//
+// ===---------------------------------------------------------------------===//
+
+#include "llvm/IR/Attributes.h"
+#include "llvm/IR/Function.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/Module.h"
+
+#include "SYCLUtils.h"
+
+namespace llvm {
+
+/// Remove a list of attributes from an IR module.
+void removeAttributes(Module &M, ArrayRef<Attribute::AttrKind> Kinds) {
+  for (auto &F : M.functions())
+    for (auto Kind : Kinds) {
+      F.removeAttributeAtIndex(AttributeList::FunctionIndex, Kind);
+      F.removeAttributeAtIndex(AttributeList::ReturnIndex, Kind);
+      for (auto &P : F.args())
+        P.removeAttr(Kind);
+      for (User *U : F.users())
+        if (CallBase *CB = dyn_cast<CallBase>(U)) {
+          CB->removeAttributeAtIndex(AttributeList::FunctionIndex, Kind);
+          CB->removeAttributeAtIndex(AttributeList::ReturnIndex, Kind);
+          for (unsigned int i = 0; i < CB->arg_size(); ++i) {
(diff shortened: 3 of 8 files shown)


clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -779,6 +781,15 @@ class SingleDeviceFunctionTracker {
     if (!CurrentDecl)
       return;
 
+    /// This will print a uniqued call graph of functions being complied for the
+    /// device. This is useful for debugging how a function ended up in device
+    /// code.
+    LLVM_DEBUG(if (!DeviceFunctions.contains(CurrentDecl)) {
+      for (std::size_t i = 0; i < CallStack.size(); i++)
```
keryell: Not clear why you have this loop.

clang/lib/Sema/SemaSYCL.cpp
```diff
@@ -779,6 +781,15 @@ class SingleDeviceFunctionTracker {
     if (!CurrentDecl)
       return;
 
+    /// This will print a uniqued call graph of functions being complied for the
```
keryell, line 771:
```suggestion
    /// This will print a uniqued call graph of functions being compiled for the
```
Can you clarify the sentence about what you mean with `uniqued` here?

llvm/lib/SYCL/SYCLUtils.cpp
```diff
@@ -0,0 +1,67 @@
+//===- SYCLUtils.cpp ------------------------------------------------------===//
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
```
keryell, line 8: Missing top level comment about the goal and content.

llvm/lib/SYCL/SYCLUtils.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include "llvm/IR/Instructions.h"
+#include "llvm/IR/IntrinsicInst.h"
+#include "llvm/IR/Intrinsics.h"
+#include "llvm/IR/IRBuilder.h"
```
keryell: Sort

llvm/lib/SYCL/SYCLUtils.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "llvm/IR/Function.h"
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/InstrTypes.h"
+#include "llvm/IR/Module.h"
+
+#include "SYCLUtils.h"
+
+namespace llvm {
+
+void removeAttributes(Module &M, ArrayRef<Attribute::AttrKind> Kinds) {
```
keryell, line 24: Add a welcoming comment on each function

llvm/lib/SYCL/VXXIRDowngrader.cpp
```diff
@@ -255,8 +218,8 @@ struct VXXIRDowngrader : public ModulePass {
                 CI->getArgOperand(0),
                 ConstantInt::getNullValue(CI->getArgOperand(0)->getType()));
             Value *Sub = B.CreateSub(
-                CI->getArgOperand(0),
-                ConstantInt::getNullValue(CI->getArgOperand(0)->getType()));
+                ConstantInt::getNullValue(CI->getArgOperand(0)->getType()),
+                CI->getArgOperand(0));
```
keryell, line 222: Curious. Was it a bug?

llvm/lib/SYCL/VXXIRDowngrader.cpp
```diff
@@ -266,10 +229,15 @@ struct VXXIRDowngrader : public ModulePass {
       I->eraseFromParent();
   }
 
-  void convertPoinsonToZero(Module &M) {
-    for (auto &PV : M.getContext().pImpl->PVConstants)
+  void convertPoisonToZero(Module &M) {
```
keryell, line 235: Comment on what and why

llvm/lib/SYCL/VXXIRDowngrader.cpp
```diff
@@ -266,10 +229,15 @@ struct VXXIRDowngrader : public ModulePass {
...
+      Type* Ty = PV.second->getType();
+      if (PV.second->use_empty() || Ty->isVoidTy() || Ty->isFunctionTy() ||
+          Ty->isLabelTy() || Ty->isMetadataTy())
+        continue;
       PV.second.get()->replaceAllUsesWith(
           Constant::getNullValue(PV.second.get()->getType()));
+    }
   }
 
   void removeMetaDataValues(Module &M) {
```
keryell, line 253: You cannot use somehow your new `removeMetadata()`?

keryell (changes requested):
Thanks for the always appreciated refactoring.

clang/lib/Sema/SemaSYCL.cpp
Ralender: By uniqued I meant that every function is printed only once even if it is called from many placed this is in place to prevent printing way too much. so it will print 1 call path by which a function was called. for example if the real call graph was:
```
A
  B
    C
      ...
  D
    C
      ...
```
this function will print:
```
A
  B
    C
      ...
  D
```

clang/lib/Sema/SemaSYCL.cpp
Ralender: This adds indentation. based on how deep in the call graph the function is.

llvm/lib/SYCL/VXXIRDowngrader.cpp
Ralender: yes. it is the bug the abs lowering I talked about in meeting.

llvm/lib/SYCL/VXXIRDowngrader.cpp
Ralender: This function removes values of type metadata because they are not supported by the HLS backend.

removeMetadata in SYCLUtils remove a specific metadata from the global scope and it is only used in AIE for now

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -60,7 +61,7 @@ struct PrepareSYCLOpt : public ModulePass {
         continue;
       }
       // Annotate kernels for HLS backend being able to identify them
-      if (F.getCallingConv() == CallingConv::SPIR_KERNEL) {
+      if (isKernelFunc(&F)) {
```
lforg37, line 86: The replacement is not equivalent, as after the first run of this pass, the kernel will no more have the SPIR_KERNEL calling convention in the HLS flow.
I don't know if there is a problem with adding the same attribute twice on the same function. 
If it is not problematic, I think the gain in clarity is worth the useless work.
Maybe using the `sycl-prepare-clearspir` opt can avoid the useless work while preserving clarity ?

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -199,7 +200,7 @@ struct PrepareSYCLOpt : public ModulePass {
   void unwrapFPGAProperties(Module &M) {
     UnwrapperVisitor UWV{};
     for (auto &F : M.functions()) {
-      if (F.getCallingConv() == CallingConv::SPIR_KERNEL) {
+      if (isKernelFunc(&F)) {
```
lforg37, line 225: Same remark here

clang/lib/Sema/SemaSYCL.cpp
keryell: Oh I see. Probably I would have used a formatting function or just 
`llvm::dbgs() << std::string { "  ", CallStack.size() };`

clang/lib/Sema/SemaSYCL.cpp
keryell: Strange... The problem is that I have no idea then that D is calling C.
So perhaps "D C" is better, displaying the second time C but no C children, because this is déjà vu. But the fact that D call C is not déjà vu.

llvm/lib/SYCL/PrepareSYCLOpt.cpp
keryell: Perhaps we could keep a kernel decoration for the non SPIR case?

llvm/lib/SYCL/VXXIRDowngrader.cpp
keryell: I just realize this is in the middle of an uncommented function, so it is difficult to catch.
Do we still need this function at the first place?
This is a strange name `lowerIntrinsic` if at the end it does handle with `std::abs` only.

llvm/lib/SYCL/VXXIRDowngrader.cpp
keryell: I guess the problem is about naming so I do not understand.
Are you talking about https://llvm.org/docs/LangRef.html#metadata-type ?
In the comment it could be clearer with
```
/// This function removes metadata type values because they are not
```

keryell (commented):
There are still some comments to be added

keryell:
@lforg37 and @Ralender need to agree on this. :-)

keryell:
Ping @Ralender

clang/lib/Sema/SemaSYCL.cpp
Ralender: it is way too noisy to be useful or performant when there isn't any uniquing. and uniquing doesn't prevent to use this for what I usually use it for, identifying why some code was generated for device.

llvm/lib/SYCL/PrepareSYCLOpt.cpp
Ralender: It is not equivalent but this shouldn't change anything to the compile flow and allow factorizing some code and expressing intent.
the kernel decoration for the non-spir case are applied by this code.

llvm/lib/SYCL/VXXIRDowngrader.cpp
Ralender: Yes this is what this function removes. to be more precise this function removes the llvm intrinsics that use metadata typed arguments. such that there is not metadata typed value in the IR.

Ralender:
This push includes the intel merge so this diff is large.

llvm/lib/SYCL/VXXIRDowngrader.cpp
keryell: Now that you have written explanation you win the right to put it in a comment ;-)

Ralender:
rebased onto sycl next

keryell (approved):
Thanks!

PR #160, opened 2022-02-15
Remove unused boost includes

mgehre-amd:
Those were not used. And now one can build sycl without having boost headers installed.


keryell (commented):
Thanks. This is probably some left-over from old way we encoded properties before using generalized variadic Clang annotations.

keryell:
Could you rebase this on `sycl/unified/next instead`? This is were we do our developments. Thanks.

keryell (approved):
Thanks for making the project cleaner!

PR #167, opened 2022-02-24
[VXX][SYCL] Fix triple normalization for fpga64

### LINKED ISSUE #166: HLS compilation flow redefines `__WINT_TYPE__` to `int`
Reproducer which shows that this breaks the standard include files:

```
$ cat foo.cpp
#include <cwchar>

$ clang++ -std=c++20 -fsycl -fsycl-targets=fpga64_hls_hw_emu -c foo.cpp
In file included from foo.cpp:1:
In file included from /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/cwchar:44:
/usr/include/wchar.h:61:22: error: typedef redefinition with different types ('unsigned int' vs 'int')
typedef unsigned int wint_t;
                     ^
/home/stepha27/sycl-workspace/sycl-build/lib/clang/14.0.0/include/stddef.h:116:23: note: previous definition is here
typedef __WINT_TYPE__ wint_t;
                      ^
1 error generated.

# Generic target for comparison
$ clang++ -std=c++20 -fsycl -c foo.cpp
```

It seems that the HLS flow redefines `__WINT_TYPE__` to `int` although it should be `unsigned int` (`bar.cpp` is empty):

```
$ clang++ -std=c++20 -fsycl -dM -E -c bar.cpp | grep WINT_TYPE
#define __WINT_TYPE__ unsigned int
#define __WINT_TYPE__ unsigned int

$ clang++ -std=c++20 -fsycl -fsycl-targets=fpga64_hls_hw_emu -dM -E -c bar.cpp | grep WINT_TYPE
#define __WINT_TYPE__ int
#define __WINT_TYPE__ unsigned int
```

The above has been produced with the current `sycl/unified/next` which was compiled by GCC 11.2.0 on a CentOS-based cluster.

Ralender:


llvm/lib/Support/Triple.cpp
```diff
@@ -1087,12 +1087,11 @@ std::string Triple::normalize(StringRef Str) {
   }
 
   if (Arch == fpga64) {
-    Components.resize(4);
+    Components.resize(std::max(Components.size(), (size_t)3));
```
keryell:
```suggestion
    Components.resize(std::max<size_t>(Components.size(), 3));
```

llvm/lib/Support/Triple.cpp
```diff
@@ -1087,12 +1087,11 @@ std::string Triple::normalize(StringRef Str) {
...
     if (Vendor == UnknownVendor)
       Components[1] = "xilinx";
-    if (Environment == UnknownEnvironment)
-      Components[3] = "sycldevice";
-    Components[2] = "unknown";
+    if (OS == UnknownOS)
+      Components[2] = "linux";
```
keryell, line 1094: What happens if this is not Linux?

Ralender: This will only add linux as a default if no OS is specified.

keryell (approved):
Thanks.

PR #178, opened 2022-04-29
Refactor parallel_for serialization and fix issue caused by new XRT

### LINKED ISSUE #175: Unnamed lambda seems not working for `parallel_for`
**Describe the bug**
Unnamed lambda seems not working for `parallel_for` when compiling. (`single_task` works.)

**To Reproduce**
Use [sycl/test/vitis/simple_tests/answer_42_unnamed.cpp](https://github.com/triSYCL/sycl/blob/sycl/unified/next/sycl/test/vitis/simple_tests/answer_42_unnamed.cpp), but replace
```c++
cgh.single_task([=] { a[0] = 42; });
```
with
```c++
cgh.parallel_for(sycl::range<1>(1), [=] (sycl::id<1> item) { a[item] = 42; });
```
Build command line: (assume the compiler is in `/opt/xilinx-llvm`)
```bash
/opt/xilinx-llvm/bin/clang -fsycl -fsycl-unnamed-lambda -std=c++20 -fsycl-targets=fpga64_hls_hw_emu answer_42_unnamed.cpp -lstdc++ -o answer_42_unnamed-xilinx-hw-emu
```
Log:
<pre>
In file included from answer_42_unnamed.cpp:1:
In file included from /opt/xilinx-llvm/bin/../include/sycl/sycl.hpp:11:
In file included from /opt/xilinx-llvm/bin/../include/sycl/CL/sycl.hpp:16:
In file included from /opt/xilinx-llvm/bin/../include/sycl/CL/sycl/backend.hpp:18:
In file included from /opt/xilinx-llvm/bin/../include/sycl/CL/sycl/detail/backend_traits_opencl.hpp:25:
In file included from /opt/xilinx-llvm/bin/../include/sycl/CL/sycl/queue.hpp:20:
<b>/opt/xilinx-llvm/bin/../include/sycl/CL/sycl/handler.hpp:1231:42: </b><font color="#C01C28"><b>error: </b></font><b>unnamed type &apos;sycl::detail::par_for_to_st&lt;(lambda at answer_42_unnamed.cpp:23:43)&gt;&apos; is invalid; provide a kernel name, or use &apos;-fsycl-unnamed-lambda&apos; to enable unnamed kernel lambdas</b>
    kernel_single_task&lt;ToSingleTaskName&gt;(SingleTaskKernel);
<font color="#26A269"><b>                                         ^</b></font>
<b>/opt/xilinx-llvm/bin/../include/sycl/CL/sycl/handler.hpp:1649:5: </b><font color="#171421"><b>note: </b></font>in instantiation of function template specialization &apos;sycl::handler::parallel_for_lambda_impl&lt;sycl::detail::auto_name, (lambda at answer_42_unnamed.cpp:23:43), 1&gt;&apos; requested here
    parallel_for_lambda_impl&lt;KernelName&gt;(NumWorkItems, std::move(KernelFunc));
<font color="#26A269"><b>    ^</b></font>
<b>answer_42_unnamed.cpp:23:11: </b><font color="#171421"><b>note: </b></font>in instantiation of function template specialization &apos;sycl::handler::parallel_for&lt;sycl::detail::auto_name, (lambda at answer_42_unnamed.cpp:23:43)&gt;&apos; requested here
      cgh.parallel_for(sycl::range&lt;1&gt;(1), [=] (sycl::id&lt;1&gt; item) { a[item] = 42; });
<font color="#26A269"><b>          ^</b></font>
1 error generated. 
</pre>

**Environment (please complete the following information):**

- OS: Ubuntu 20.04 / 22.04
- Target device and vendor: xilinx_u200_gen3x16_xdma_1_202110_1
- DPC++ version: `clang version 15.0.0 (git@github.com:triSYCL/sycl.git 29df7b7d84121137b5ec0f370cf24b2ccfe3683f)` (current sycl/unified/next branch) with Vitis 2021.2
- Dependencies version: XRT build version: 2.14.0

**Additional context**
* On Ubuntu 20.04 the `parallel_for` one doesn't compile; the `single_task` one runs fine
* ~~On Ubuntu 22.04 the `parallel_for` one doesn't compile; the `single_task` one says `No kernel named E_clES2_E10fourty_two_8rZJhPCI was found -46 (CL_INVALID_KERNEL_NAME)` at runtime. 
  Still trying to configure my environment as the latter one seems not related to this issue.~~
  Not related to system version, but XRT version. (See below)

Ralender:


Ralender:
I will add a test for this.

fxzjshm:
Sorry if I disturb you, but I have some problem using this branch.  
The current implementation requires the kernel to be device copyable, which shouldn't be required (I think, at least non-device-copyable kernels compiles on other SYCL implementations.)  

Consider this situation: a kernel `K` uses a parameter `P` that isn't device copyable because it isn't "TriviallyCopyable" (in my case, `P`'s type is `boost::iterators::counting_iterator<T>`), but I think it should work on device, so I explicitly added `
template <typename T> struct sycl::is_device_copyable<boost::iterators::counting_iterator<T>> : std::true_type {};`, and it really works.
However, this parameter `P` makes the kernel `K` itself non-"TriviallyCopyable" and non-device-copyable, which is required in this implemention (maybe by `detail::serialize_parallel_for()` [here](https://github.com/triSYCL/sycl/blob/1a5879799992020634aaa099e47e34f0adbe12ea/sycl/include/CL/sycl/handler.hpp#L1116) and [here](https://github.com/triSYCL/sycl/blob/1a5879799992020634aaa099e47e34f0adbe12ea/sycl/include/CL/sycl/handler.hpp#L1747) as the kernel `K` becomes `detail::serialize_parallel_for()`'s parameter). In my humble opinion, It's hard to deal with this using explicit `sycl::is_device_copyable<...>` as the kernel is often a lambda.  

Could you help me how to deal with this problem?

Ralender:
I pushed a commit that should address you issue. becasue we should respect user provided specializations of `sycl::is_device_copyable`.
I also noticed that it doesn't seem like there is any checking that the lambda only contains device copyable type. for parallel_for.

but it is rarely safe to bitwise copy to the device a type that are not `TriviallyCopyable`. especially an iterator which is likely to contain a pointer. I didn't look into `boost::iterators::counting_iterator` to check if it would be safe.

sycl/tools/sycl-vxx/bin/sycl_vxx_post_link.py
```diff
@@ -18,12 +18,14 @@ def extract(self):
...
             process = subprocess.run(["xclbinutil", "-i", xclbin,
                                       "--info"], capture_output=True)
+            klist = []
             for line in process.stdout.decode("utf8").split("\n"):
-                if line.strip().startswith("Kernels:"):
-                    line = line.replace("Kernels:", "", 1).strip()
-                    kernels = {s.strip() for s in line.split(',')}
-                    return kernels
-            raise EOFError("No kernel found")
+                if line.strip().startswith("Kernel:"):
```
keryell, line 26: Add somewhere an example of the output we are trying to parse

fxzjshm:
Yes, in most cases you are right. But in some edge cases it just don't work 😢 
(e.g. `std::pair`-like user-defined templated class with custom copy assignment operator, even if that `operator=` does the same thing as `= default` when `T` is TriviallyCopyable)

Anyway, thanks for this!

PR #179, opened 2022-05-03
[VitisIP] Fixes for archgenlib

Ralender:


clang/lib/Frontend/InitPreprocessor.cpp
```diff
@@ -1294,6 +1294,10 @@ static void InitializePredefinedMacros(const TargetInfo &TI,
     Builder.defineMacro(
         "__VITIS_KERNEL",
         "extern \"C\" __attribute__((annotate(\"vitis_kernel\")))");
+  else
+    /// This is needed to compile ArchGenLib but I think this else case is not
+    /// the correct way to address the issue
```
keryell: Why we have this else case, by the way?

keryell (commented):
It is not clear what is done here and why.

clang/lib/Frontend/InitPreprocessor.cpp
Ralender: We experimented with this while using the VitisIP flow on archgen lib. and it did remove some preprocessor in some simple cases. but it prevents from detecting what is the compilation mode. which is a lot worse.

keryell (approved):
Thanks!

PR #194, opened 2022-08-12
Improve for 2022.1

keryell:
Mainly using a part for Vitis IP which is installed by default.

### DIFF
--- a/sycl/doc/GettingStartedXilinxFPGA.md
+++ b/sycl/doc/GettingStartedXilinxFPGA.md
@@ -915,7 +915,7 @@ treating compiler warnings as errors and producing a compiler database
 to be used by tools like LSP server like `clangd`:
 
 ```bash
-python3 $SYCL_HOME/llvm/buildbot/configure.py --cuda \
+python3 $SYCL_HOME/llvm/buildbot/configure.py --xrt --cuda \
   --cmake-opt="-DCMAKE_EXPORT_COMPILE_COMMANDS=1"
 ```
 
--- a/sycl/doc/GettingStartedXilinxFPGA.md
+++ b/sycl/doc/GettingStartedXilinxFPGA.md
@@ -910,12 +910,12 @@ python3 $SYCL_HOME/llvm/buildbot/compile.py
 ```
 
 These scripts have many options which can be displayed when using the
-`--help` option. For example to configure with CUDA support, without
-treating compiler warnings as errors and producing a compiler database
-to be used by tools like LSP server like `clangd`:
+`--help` option. For example to configure with CUDA support, shared
+libraries and producing a compiler database to be used by tools like
+LSP server like `clangd`:
 
 ```bash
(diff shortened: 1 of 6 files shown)


sycl/test/vitis/simple_tests/single_task_vector_add.cpp
```diff
@@ -4,7 +4,7 @@
 // RUN: %clangxx -std=c++20 -fsycl -fsycl-targets=%sycl_triple %s -o %t.dir/exec.out
 // RUN: %ACC_RUN_PLACEHOLDER %t.dir/exec.out
 // Test that we can get a fast reports by adding --vitis-ip-part
-// RUN: %clangxx -std=c++20 -fsycl -fsycl-targets=%sycl_triple %s -o %t.dir/exec2.out --vitis-ip-part=xc7vx330t-ffg1157-1
+// RUN: %clangxx -std=c++20 -fsycl -fsycl-targets=%sycl_triple %s -o %t.dir/exec2.out --vitis-ip-part=xcu200-fsgd2104-2-e
```
Ralender, line 7: What we should do here is figure out inside lit what part id is available on the system and setup a substitution for it. similarly to what we do for "%opencv4_flags"
not everyone will have an u200

sycl/doc/GettingStartedXilinxFPGA.md
```diff
@@ -915,7 +915,7 @@ treating compiler warnings as errors and producing a compiler database
 to be used by tools like LSP server like `clangd`:
 
 ```bash
-python3 $SYCL_HOME/llvm/buildbot/configure.py --cuda \
+python3 $SYCL_HOME/llvm/buildbot/configure.py --xrt --cuda \
```
Ralender:
```suggestion
python3 $SYCL_HOME/llvm/buildbot/configure.py --xrt --cuda --shared-libs \
```
shared libs greatly speed up the build so Ithink it should be added by default

Ralender (commented):
Seems good, but I added some suggestion for improvement

sycl/test/vitis/simple_tests/single_task_vector_add.cpp
keryell: I think it is another problem.
Not everybody has an Alveo U200 platform installed but the `xcu200-fsgd2104-2-e` part comes de facto with Vivado.
But anyway, having the part should be configurable too.

sycl/doc/GettingStartedXilinxFPGA.md
keryell: Good point! I thought it was the case by default.

sycl/test/vitis/simple_tests/single_task_vector_add.cpp
Ralender: if xcu200-fsgd2104-2-e comes by default with Vitis/Vivado its fine keeping it.

Ralender (approved):

PR #195, opened 2022-09-01
Fixes for archgen-mlir

Ralender:


llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -230,12 +231,48 @@ struct PrepareSYCLOptState {
     }
   }
 
+  void removeCallInstCasts(llvm::Module *M) {
```
keryell, line 237: This PR is for a fix, so it might not have been obvious at the first place. :-)
What about a few comments on what is the problem, what does it does and why ?...

Ralender: I added a few comments

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -230,12 +231,55 @@ struct PrepareSYCLOptState {
     }
   }
 
+  /// transform call on casted functions into call on functions with casted
+  /// arguments. because calls on casted function are not inlined and the Vitis
+  /// backend has issues with call on casted functions
```
keryell:
```suggestion
  /// Transform calls on casted functions into calls on functions with casted
  /// arguments, because calls on casted functions are not inlined and the Vitis
  /// backend has issues with call on casted functions.
```

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -230,12 +231,55 @@ struct PrepareSYCLOptState {
...
+            if (auto *CB = dyn_cast<CallBase>(U))
+              if (CB->use_empty() &&
+                  llvm::all_of(F.getFunctionType()->params(),
+                               [](Type *Ty) { return Ty->isPointerTy(); }) &&
+                  llvm::all_of(CB->operand_values(), [](Value *v) {
+                    return v->getType()->isPointerTy();
+                  })) {
+                SmallVector<Value *> args;
+
+                /// emit castes for all arguments that need them
```
keryell:
```suggestion
                /// Emit casts for all arguments that need them
```

llvm/lib/SYCL/PrepareSYCLOpt.cpp
```diff
@@ -230,12 +231,55 @@ struct PrepareSYCLOptState {
...
+                    continue;
+                  }
+                  auto *cast = llvm::BitCastInst::CreatePointerCast(
+                      CB->getOperand(idx),
+                      F.getFunctionType()->getParamType(idx));
+                  cast->insertBefore(CB);
+                  args.push_back(cast);
+                }
+
+                /// replace the call
```
keryell:
```suggestion
                /// Replace the call
```

keryell (changes requested):
More comments would be appreciated. :-)

keryell:
You can configure your Git tool with a spell-checker and fix the message of [Fix mempy unrollig](https://github.com/triSYCL/sycl/pull/195/commits/a3f0b8586ec11268d2558da4b4151095c704c8bf)

PR #213, opened 2022-12-14
[SYCL][Driver] do not generate device (for FPGA) code with PIC

Ralender:
PIC has no impact on the device side

also add documentation about an issue that can occure when using dynamic librairies


clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -5700,7 +5700,7 @@ void Clang::ConstructJob(Compilation &C, const JobAction &JA,
     CmdArgs.push_back("-mrelocation-model");
     CmdArgs.push_back(RMName);
   }
-  if (PICLevel > 0) {
+  if (PICLevel > 0 && !Triple.isXilinxFPGA()) {
```
keryell, line 5704: Explain why

keryell (approved):

keryell:
Thanks for the fix and clarification of the documentation.

PR #216, opened 2022-12-28
Merge from upstream

Ralender:

PR #218, opened 2023-01-01
PLRAM support

bsobol:
This adds support for PLRAM memory 'banks'. In most cases it is just cloning the implementation of DDR/HBM with minor adjustments.

I agree that `sycl/include/sycl/ext/xilinx/fpga/memory_properties.hpp` deserves some refactoring (and even more with plram case), but I didn't try to do that in this PR as I don't know if you have specific plans for that.

### DIFF
--- a/llvm/lib/SYCL/SYCLUtils.cpp
+++ b/llvm/lib/SYCL/SYCLUtils.cpp
@@ -117,6 +117,8 @@ constexpr const char *xilinx_ddr_bank =
     "sycl_xilinx_ddr_bank";
 constexpr const char *xilinx_hbm_bank =
     "sycl_xilinx_hbm_bank";
+constexpr const char *xilinx_plram_bank =
+    "sycl_xilinx_plram_bank";
 
 /// getAttributeAtIndex(0, ...) is the attribute on the return. The first argument
 /// starts at 1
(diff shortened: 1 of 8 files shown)


llvm/lib/SYCL/SYCLUtils.cpp
```diff
@@ -191,16 +193,23 @@ void giveNameToArguments(Function &F) {
 }
 
 void annotateMemoryBank(Argument *Arg, MemBankSpec Bank) {
+  const auto mem_type = [&Bank](){
```
keryell: Be lazy :-)
```suggestion
  const auto mem_type = [&Bank] {
```

keryell (commented):
Cool ! :-)
Would it possible to add a test using it?
Thanks.

Ralender:
seems good to me. except for the tests.
you cant use https://github.com/triSYCL/sycl/blob/sycl/unified/master/sycl/test/vitis/simple_tests/ddr_bank_test.cpp
as an example of how ddr and hbm banks are tested.

Ralender (approved):
Thank you.

sycl/test/vitis/simple_tests/memory_types_test.cpp
```diff
@@ -0,0 +1,43 @@
+// REQUIRES: vitis
+
+// RUN: rm -rf %t.dir && mkdir %t.dir && cd %t.dir
+// RUN: %clangxx %EXTRA_COMPILE_FLAGS-fsycl -std=c++20 -fsycl-targets=%sycl_triple %s -o %t.dir/exec.out > %t.check 2>&1
+// RUN: %run_if_not_cpu FileCheck --input-file=%t.check %s
+// RUN: %ACC_RUN_PLACEHOLDER %t.dir/exec.out
+
+#include <sycl.hpp>
```
keryell:
```suggestion
#include <sycl/sycl.hpp>
```
@Ralender curious. I think upstream was changed to avoid such a thing to compile. Perhaps a bad merge conflict? Or perhaps it is already fixed in https://github.com/triSYCL/sycl/pull/216 ?

sycl/test/vitis/simple_tests/memory_types_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+  sycl::buffer<int, 1> BufferA{4}, BufferB{4}, BufferC{4};
+  sycl::queue Queue{};
+  const std::size_t buf_size{BufferA.size()};
+
+  Queue.submit([&](sycl::handler &cgh) {
+    sycl::ext::oneapi::accessor_property_list PL{sycl::ext::xilinx::ddr_bank<1>};
+    // CHECK-DAG:{{.*}}:DDR[1]
+    sycl::accessor Accessor(BufferA, cgh, sycl::write_only, PL);
+    cgh.single_task<class SmallerTesta>([=]{
+          for (std::size_t i = 0 ; i < buf_size ; ++i)
```
keryell: You could use `std::iota`

keryell (commented):
Great contribution.

sycl/test/vitis/simple_tests/memory_types_test.cpp
bsobol: done

keryell (approved):
Thanks!

sycl/test/vitis/simple_tests/memory_types_test.cpp
Ralender: There is still 300+ uses of 
`#include <sycl/sycl.hpp>`
so I dont think there as a serious attempt are avoid it.

keryell: Curious, I think I have seen a removal of `#include <sycl.hpp>` upstream with a CMake change to avoid such examples to compile.

Ralender: oh I mis-undrsetood what you meant.
`#include <sycl.hpp>`
is indeed barely used anymore.

PR #223, opened 2023-04-03
Merge upstream

Ralender:
merge with the latest(at the time of starting) version of intel/sycl.
then fix failing tests


keryell (approved):
Thanks.

Issue #224, opened 2023-04-07
`xrt_piEventGetProfilingInfo` not implemented


keryell:
The example from https://github.com/triSYCL/sycl/blob/sycl/unified/next/sycl/doc/GettingStartedXilinxFPGA.md#running-a-bigger-example-on-real-fpga is using
```c++
  queue q {property::queue::enable_profiling()};
```
and fails with
```
keryell@xsjsycl41:/var/tmp/rkeryell/SYCL/llvm/sycl/test/vitis/edge_detection ((03ee1150c1c7...))$ ./edge_detection data/input/eiffel.bmp
Calculating Max Energy... 
inputBits = 8 coefMax = 2 
Max Energy = 14 Bits 
Image Dimensions: [1024 x 1895]
Used Size: 1895x1024 = 1940480
Launching Kernel... 
pixel_rb size in submit: 1940480 
pixel_rb count in submit: 1940480 
Getting Result... 
pi_xrt: unimplemented: call to unimplemented function in XRT backend: pi_result xrt_piEventGetProfilingInfo(pi_event, pi_profiling_info, size_t, void*, size_t*)
terminate called without an active exception
Aborted
```

PR #226, opened 2023-04-12
[SYCL][XRT] Add partial profilling support to pi_xrt

Ralender:
This changes adds basic support for profiling. this should fix https://github.com/triSYCL/sycl/issues/224

### DIFF
--- a/sycl/plugins/xrt/pi_xrt.cpp
+++ b/sycl/plugins/xrt/pi_xrt.cpp
@@ -35,23 +35,20 @@
 #include <sycl/detail/pi.h>
 #include <sycl/detail/pi.hpp>
 
-#include <array>
 #include <atomic>
 #include <cassert>
+#include <chrono>
 #include <cstddef>
 #include <cstdlib>
 #include <cstring>
 #include <deque>
 #include <functional>
-#include <iterator>
 #include <limits>
 #include <mutex>
 #include <numeric>
-#include <optional>
 #include <ostream>
 #include <stdint.h>
 #include <string>
-#include <thread>
 #include <type_traits>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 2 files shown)


sycl/plugins/xrt/pi_xrt.cpp
```diff
@@ -35,26 +35,23 @@
...
 #include <numeric>
-#include <optional>
 #include <ostream>
 #include <stdint.h>
 #include <string>
-#include <thread>
 #include <type_traits>
 #include <utility>
 #include <vector>
+#include <chrono>
```
keryell: Sort.

sycl/plugins/xrt/pi_xrt.cpp
```diff
@@ -1325,9 +1364,23 @@ pi_result xrt_piQueueCreate(pi_context context, pi_device device,
...
   assert_valid_obj(device);
-  // TODO(XRT): : properties not handled
-
-  *queue = make_ref_counted<_pi_queue>(context, device).give_externally();
+  if (properties) {
+    if (properties & PI_QUEUE_FLAG_PROFILING_ENABLE)
+      std::cerr
+          << "warning: support for profiling is only partial. pi_xrt is fully "
+             "synchonous so it will detect when an event is completed only "
+             "when the user waits on the event. if the user waits on the event "
```
keryell:
```suggestion
             "synchronous so it will detect when an event is completed only "
             "when the user waits for the event. If the user waits on the event "
```

sycl/plugins/xrt/pi_xrt.cpp
```diff
@@ -1325,9 +1364,23 @@ pi_result xrt_piQueueCreate(pi_context context, pi_device device,
...
+             "just after starting it, the profiling information is reliable. "
+             "but if the user uses wait on an event that was already finished "
+             "some time ago, pi_xrt will notice it is finished at the time of "
+             "the wait not at the prior 'real' time the event was completed"
```
keryell:
```suggestion
             "but if the user uses wait for an event that was already finished "
             "some time ago, pi_xrt will notice it is finished at the time of "
             "the wait not at the prior 'real' time the event was completed"
```

keryell (approved):

PR #234, opened 2023-08-15
[SYCL][AIE][DOC] Improve documentation

keryell:
Minor layout fixes, rename some Xilinx to AMD.


README.md
```diff
@@ -37,6 +37,21 @@ driver and scripts to use AMD Vitis
...
+for mathematical fixed-point functions to optimize hardware usage for
+low precision computations. This was presented at the following conference:
+
+>  Luc FORGET, Gauthier HARNISCH, Ronan KERYELL and Florent DE
+>  DINECHIN. « A single-source C++ 20 HLS ﬂow for function evaluation
+>  on FPGA and beyond. » *In HEART2022: International Symposium on
+>  Highly-Efficient Accelerators and Reconﬁgurable Technologies*, pages
+>  51–58. Association for Computing Machinery, Tsukuba, Japan,
+>  June 2022. doi:10.1145/3535044. 3535051. https://hal.archives-ouvertes.fr/hal-03684757
+
```
Ralender, line 55: What we presented was ArchGen not ArchGenMLIR.
ArchGenMLIR is almost the same API but built with a better design on top of MLIR.

sycl/doc/AMDGettingStartedAIE.md
```diff
@@ -6,61 +6,66 @@ change rapidly or get outdated.
...
-executed on the host). This makes it superb for debugging your applications 
-before trying to target your application on a more complex heterogeneous 
+device compiler and defaults to a multi-threaded CPU application (all kernels are
+executed on the host). This makes it superb for debugging your applications
+before trying to target your application on a more complex heterogeneous
 architecture like AIE.
 
-The AIE++ library for now only targets AIE hardware.
+The AIE++ library for now only targets AIE hardware as the CPU
+implementation is a work-in-progress.
```
Ralender, line 46: we can already do quite many thing with AIE++ on cpu if I remember correctly.
it might be nice to show the difference in APIs between the two. since the better API is part of why AIE++ was built.

keryell: Indeed. But since I have not documented it yet, my plan was to add this at the same time.
Nice to see you again in the area. ;-)

